PR #47, opened 2026-06-16
feat(mcp): add protected HTTP MCP server support

frederikprijck:
## Summary

Adds support for **protected HTTP MCP servers** — servers that require an `Authorization: Bearer` token rather than being publicly reachable (e.g. the Auth0 hosted MCP server, which authenticates with a Management API token).

This is the auth-plumbing layer only. The `calledTool`/`calledToolOneOf` trace-based grader primitives and the `hosted_mcp_list_applications` eval that build on top of it are intentionally left for a follow-up PR.

## What's included

- **`mintMcpToken`** — per-job client-credentials token mint for HTTP MCP servers (minted per job so a long matrix run never reuses an expired token).
- **`MCPHttpServerConfig.auth`** — new optional `auth` block (`tokenUrl` / `clientId` / `clientSecret` / `audience`) typed as `MCPOAuthConfig`.
- **All four runners forward the token**: `claude-code`, `copilot`, and `gemini-cli` set an `Authorization: Bearer` header on the MCP server config; `codex` writes a `bearer_token_env_var` reference into `config.toml` and injects the token into its process env (an inline `bearer_token` is rejected by Codex, so the secret stays out of the file).
- **Loud failure** — a failed token mint skips the server with a `logger.warn` rather than registering it unauthenticated.
- **`sandbox.passthroughEnv`** — forward named host env vars into the Docker sandbox (names only; values resolved from `process.env` at launch) so MCP credentials reach the container.
- **`docs/PROTECTED_MCP.md`** — setup guide (prerequisites, env vars, config, how minting/forwarding works, troubleshooting).

## Test plan

- [x] `npm run build` passes
- [x] `npm test` passes (697 tests)
- [x] `npm run lint` passes
- [ ] Manual: run an eval with `--tools mcp` against a protected server with `MCP_TENANT_DOMAIN`/`MCP_CLIENT_ID`/`MCP_CLIENT_SECRET` set and confirm the token is minted and forwarded for each runner

## Summary by CodeRabbit

* **New Features**
  * Added support for protected MCP servers with per-job OAuth token minting.
  * Docker sandbox jobs can now receive selected host environment variables.
  * Several runners now forward authenticated MCP access automatically.

* **Bug Fixes**
  * Authenticated MCP servers are skipped with a warning if token minting fails, instead of being registered without credentials.
  * Improved handling for missing or incomplete auth settings.

* **Documentation**
  * Added guidance for configuring protected MCP servers and troubleshooting common setup issues.

### DIFF
--- a/packages/eval-core/src/config/mcp-auth.ts
+++ b/packages/eval-core/src/config/mcp-auth.ts
@@ -8,6 +8,9 @@
 import type { MCPOAuthConfig } from './framework.js';
 import { logger } from '../utils/logger.js';
 
+/** Fail-fast budget for the token endpoint — a hung endpoint must not block job startup. */
+const TOKEN_REQUEST_TIMEOUT_MS = 10_000;
+
 /**
  * Derives the environment-variable name a runner uses to pass a minted Bearer
  * token to an MCP server, keeping the secret out of on-disk config files.
@@ -42,23 +45,39 @@ export async function mintMcpToken(auth: MCPOAuthConfig): Promise<string | undef
         client_secret: auth.clientSecret,
         audience: auth.audience,
       }),
+      // Fail fast: a token endpoint that accepts the connection but never
+      // responds would otherwise block job startup until the 30-min task
+      // timeout. AbortSignal.timeout rejects the fetch after 10s instead.
+      signal: AbortSignal.timeout(TOKEN_REQUEST_TIMEOUT_MS),
     });
+    // Read the raw body once so it's preserved for diagnosis whether the
+    // status is an error OR a 200 with a non-JSON payload (e.g. a proxy/gateway
+    // HTML error page) — res.json() would consume it and leave nothing to log.
+    const body = await res.text().catch(() => '');
     if (!res.ok) {
-      // Include the response body — for the documented access_denied case it
-      // carries the error/error_description, making misconfig far easier to
-      // diagnose.
-      const body = await res.text().catch(() => '');
+      // For the documented access_denied case the body carries the
+      // error/error_description, making misconfig far easier to diagnose.
       logger.warn(`[mcp-auth] Token request failed: ${res.status}${body ? ` — ${body}` : ''}`);
       return undefined;
     }
-    const { access_token } = (await res.json()) as { access_token?: string };
+    let access_token: string | undefined;
+    try {
+      ({ access_token } = JSON.parse(body) as { access_token?: string });
+    } catch {
+      logger.warn(`[mcp-auth] Token response was not valid JSON${body ? ` — ${body}` : ''}`);
+      return undefined;
+    }
     if (!access_token) {
       logger.warn('[mcp-auth] Token response missing access_token');
       return undefined;
     }
     return access_token;
   } catch (err) {
-    logger.warn(`[mcp-auth] Token request error: ${err instanceof Error ? err.message : String(err)}`);
+    // Include the token endpoint so a DNS/TLS/network failure is
+    // distinguishable from a misconfigured URL.
+    logger.warn(
+      `[mcp-auth] Token request error for ${auth.tokenUrl}: ${err instanceof Error ? err.message : String(err)}`,
+    );
     return undefined;
   }
 }
--- a/packages/eval-core/src/config/mcp-auth.ts
+++ b/packages/eval-core/src/config/mcp-auth.ts
@@ -8,16 +8,35 @@
 import type { MCPOAuthConfig } from './framework.js';
 import { logger } from '../utils/logger.js';
 
+/**
+ * Derives the environment-variable name a runner uses to pass a minted Bearer
+ * token to an MCP server, keeping the secret out of on-disk config files.
+ * Runners reference this name indirectly — Codex via `bearer_token_env_var` in
+ * config.toml, Gemini CLI via `$VAR` expansion in settings.json.
+ *
+ * Assumes server names are distinct after normalization: uppercasing and
+ * mapping non-alphanumerics to `_` means names like `auth0-hosted` and
+ * `auth0.hosted` would collide to the same env var. Config keys are authored by
+ * hand and never differ only by punctuation, so this holds.
+ */
+export function mcpBearerTokenEnvVar(serverName: string): string {
+  return `MCP_BEARER_${serverName.replace(/[^A-Za-z0-9]/g, '_').toUpperCase()}`;
+}
+
 export async function mintMcpToken(auth: MCPOAuthConfig): Promise<string | undefined> {
   if (!auth.tokenUrl || !auth.clientId || !auth.clientSecret || !auth.audience) {
     logger.warn('[mcp-auth] Incomplete OAuth config — skipping token mint');
     return undefined;
   }
   try {
+    // Form-encode the request: RFC 6749 §4.4.2 mandates
+    // application/x-www-form-urlencoded for the client-credentials grant, so
+    // this is the most broadly compatible format for an arbitrary protected
+    // MCP server's token endpoint. Auth0's /oauth/token accepts it too.
     const res = await fetch(auth.tokenUrl, {
       method: 'POST',
-      headers: { 'content-type': 'application/json' },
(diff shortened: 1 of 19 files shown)


packages/eval/src/sandbox/docker.ts
```diff
@@ -150,6 +155,15 @@ export async function runJobInDocker(options: DockerRunOptions): Promise<JobResu
     envFlags.push('-e', `GH_TOKEN=${ghToken}`);
   }
 
+  // Forward app-declared passthrough env vars (e.g. MCP server credentials).
+  // Only vars currently set on the host are forwarded; missing ones are skipped.
+  for (const name of passthroughEnv ?? []) {
```
sanchitmehtagit, line 161: Since this exists specifically to carry MCP credentials, a quick log of which names were forwarded vs. skipped would make the most common failure — "I forgot to `export MCP_CLIENT_SECRET`" — obvious from the run output instead of surfacing later as a mysterious token-mint failure inside the container. Totally optional.

packages/eval/src/runners/claude-code/agent.ts
```diff
@@ -137,12 +138,25 @@ export async function runClaudeCodeAgent(
...
     const configServers = getFrameworkConfig().mcp.servers;
-    const httpServers: Record<string, { type: 'http'; url: string }> = {};
+    const httpServers: Record<string, { type: 'http'; url: string; headers?: Record<string, string> }> = {};
     for (const [name, server] of Object.entries(configServers)) {
-      if (server.type === 'http') {
+      if (server.type !== 'http') continue;
+      if (server.auth) {
+        const token = await mintMcpToken(server.auth);
+        if (!token) {
+          logger.warn(`[ClaudeCode] MCP server '${name}' skipped — token mint failed or creds missing`);
```
sanchitmehtagit, line 151: This auth path (mint → inject header, skip on failure) has no test on the claude-code runner — codex, copilot, and gemini each got success + mint-failure cases, but `claude-code-agent.test.ts` has none. Since this is the default runner for `claude-*` models, worth adding the same trio: authed server → `Authorization` header injected; mint fails → server skipped; unauthed → no header.

(Anchored here on the source path since the test file has no changes in this PR — there's no diff line to attach to over there.)

packages/eval/src/runners/claude-code/agent.ts
```diff
@@ -137,12 +138,25 @@ export async function runClaudeCodeAgent(
   }
 
   // Build MCP server config when --tools mcp is requested.
-  let mcpServers: Record<string, { type: 'http'; url: string }> | undefined;
+  // Token is minted here (job start) so a long matrix run never reuses an expired token.
+  let mcpServers: Record<string, { type: 'http'; url: string; headers?: Record<string, string> }> | undefined;
```
sanchitmehtagit, line 142: Minor observability gap (relates to the log line further down that reads `if (mcpServers) logger.info(...)`): when *all* servers fail to mint, `mcpServers` stays undefined and nothing is logged — so an all-fail run reads identically to "MCP was never requested." Consider logging when MCP was requested but ended up with zero available servers, and/or recording "MCP requested but unavailable" into the run record so a misconfig surfaces in results rather than only in an early warn. (Not asking to make it fatal — graceful degradation across a big matrix is the right call.)

packages/eval-core/src/config/mcp-auth.ts
```diff
@@ -0,0 +1,49 @@
...
+    });
+    if (!res.ok) {
+      // Include the response body — for the documented access_denied case it
+      // carries the error/error_description, making misconfig far easier to
+      // diagnose.
+      const body = await res.text().catch(() => '');
+      logger.warn(`[mcp-auth] Token request failed: ${res.status}${body ? ` — ${body}` : ''}`);
+      return undefined;
+    }
+    const { access_token } = (await res.json()) as { access_token?: string };
```
sanchitmehtagit: Nice-to-have robustness: if the endpoint returns 200 with a non-JSON body (e.g. a proxy/gateway HTML error page), `res.json()` throws and gets caught below as a generic "Token request error" — with the body already consumed, so it's not logged. Reading `res.text()` first and then `JSON.parse` would preserve the body for diagnosis, same as the `!res.ok` branch above already does.

packages/eval-core/src/config/mcp-auth.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (!access_token) {
+      logger.warn('[mcp-auth] Token response missing access_token');
+      return undefined;
+    }
+    return access_token;
+  } catch (err) {
```
sanchitmehtagit, line 75: Two small hardening ideas here (both non-blocking):

1. This catch logs `err.message` only — for a DNS/TLS/network failure it's hard to tell "config is wrong" from "network is down." Including `auth.tokenUrl` (and maybe the stack) would help.
2. There's no fetch timeout — a hung token endpoint blocks until the 30-min job timeout. An `AbortSignal.timeout(...)` on the `fetch` would fail fast.

packages/eval/src/runners/codex/agent.ts
```diff
@@ -433,11 +449,40 @@ export async function runCodexAgent(
...
+      const token = await mintMcpToken(server.auth);
+      if (!token) {
+        logger.warn(`[Codex] MCP server '${name}' skipped — token mint failed or creds missing`);
+        continue;
+      }
+      // Assumes server names are distinct after normalization: uppercasing and
+      // mapping non-alphanumerics to `_` means names like `auth0-hosted` and
+      // `auth0.hosted` would collide to the same env var. Config keys are
+      // authored by hand and never differ only by punctuation, so this holds.
+      const envVar = `MCP_BEARER_${name.replace(/[^A-Za-z0-9]/g, '_').toUpperCase()}`;
```
sanchitmehtagit: The comment here is honest about the assumption, and it holds for today's hand-authored configs — so this is just future-proofing, not a bug. `auth0-hosted` and `auth0.hosted` both normalize to `MCP_BEARER_AUTH0_HOSTED`, and the second would silently clobber the first's token. A one-liner would remove the footgun entirely: throw if the normalized name is already a key in `bearerTokens`. Cheap insurance against a confusing future debugging session.

packages/eval/src/sandbox/docker.ts
frederikprijck: Done in ce6b49f — docker now logs forwarded vs skipped passthrough env **names** (never values), so a forgotten `export MCP_CLIENT_SECRET` surfaces up front.

packages/eval/src/runners/claude-code/agent.ts
frederikprijck: Done in ce6b49f — added the trio to `claude-code-agent.test.ts`: authed→`Authorization` header injected, mint-fail→server skipped + warn, unauthed→no header.

packages/eval/src/runners/claude-code/agent.ts
frederikprijck: Done in ce6b49f — when `--tools mcp` is requested but no server becomes available, every runner now logs a `--tools mcp requested but no MCP servers are available` warning.

packages/eval-core/src/config/mcp-auth.ts
frederikprijck: Done in ce6b49f — the raw body is now read once via `res.text()` then `JSON.parse`d, so a 200 with a non-JSON payload is preserved and logged for diagnosis.

packages/eval-core/src/config/mcp-auth.ts
frederikprijck: Done in ce6b49f — the catch now includes `auth.tokenUrl` in the log, and the fetch carries `AbortSignal.timeout(10_000)` so a hung endpoint fails fast.

packages/eval/src/runners/codex/agent.ts
frederikprijck: Done in ce6b49f.

sanchitmehtagit (approved):

PR #55, opened 2026-06-19
fix(compile): auto-run npm install before compile_command

sanchitmehtagit:
## Summary

- `runCompileCommand` now prepends `npm install` before the eval's `compile_command` whenever a `package.json` exists in the workspace
- Fixes builds failing with "module not found" when the agent added new dependencies to `package.json` but never ran `npm install` itself
- Two new tests: one verifying auto-install is prepended with a `package.json` present, one verifying it's skipped when absent

## Test plan

- [ ] `npm run build && npm test` passes
- [ ] Run an eval where the agent adds a new dep to `package.json` without running `npm install` — `compiles` grader should now pass

### DIFF
--- a/packages/eval-core/src/workspace/workspace.ts
+++ b/packages/eval-core/src/workspace/workspace.ts
@@ -99,6 +99,8 @@ export interface RunSetupCommandOptions {
 export interface RunCompileCommandOptions {
   /** Timeout in ms for the compile command. Defaults to {@link DEFAULT_FRAMEWORK_CONFIG}.workspace.compileCommandTimeoutMs. */
   timeoutMs?: number;
+  /** When provided, this command is prepended to the compile command sequence (e.g. the eval's setup_command to ensure deps are installed). */
+  setupCommand?: string;
 }
 
 /**
@@ -191,10 +193,8 @@ export function runCompileCommand(
     return { ...base, output: `compile command has an empty segment: ${command}` };
   }
 
-  // Auto-install dependencies before building so new packages the agent added
-  // to package.json are available even if the agent never ran npm install itself.
-  if (existsSync(join(workspace, 'package.json'))) {
-    subCommands.unshift('npm install');
+  if (options?.setupCommand) {
+    subCommands.unshift(options.setupCommand);
   }
 
   let combinedOutput = '';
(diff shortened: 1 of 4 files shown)


sanchitmehtagit:
This PR does not solve the problem when Agents add malicious dependencies. and we install them as part of compile judge

packages/eval-core/src/workspace/workspace.ts
```diff
@@ -191,6 +191,12 @@ export function runCompileCommand(
     return { ...base, output: `compile command has an empty segment: ${command}` };
   }
 
+  // Auto-install dependencies before building so new packages the agent added
+  // to package.json are available even if the agent never ran npm install itself.
+  if (existsSync(join(workspace, 'package.json'))) {
+    subCommands.unshift('npm install');
+  }
```
frederikprijck: Can we use the existing setup_command here, and only unshift it when it exists.

sanchitmehtagit: yes

sanchitmehtagit: Done in 5c95c3a — added `setupCommand?` to `RunCompileCommandOptions` and wired callers to pass `evalDef.setupCommand`; removed the `package.json` heuristic.

packages/eval-core/src/workspace/workspace.ts
```diff
@@ -191,6 +193,10 @@ export function runCompileCommand(
     return { ...base, output: `compile command has an empty segment: ${command}` };
   }
 
+  if (options?.setupCommand) {
+    subCommands.unshift(options.setupCommand);
+  }
+
```
frederikprijck, line 202:
```suggestion
  // When a setupCommand is provided, we ensure to run it before verifying compilation. 
  // This is done because there is no guarantee that an agent did not add a dependency 
  // without running the installation command.
  if (options?.setupCommand) {
    subCommands.unshift(options.setupCommand);
  }
```

sanchitmehtagit: Done in 1459bf3 — added the suggested comment block.

frederikprijck (approved):

PR #56, opened 2026-06-19
feat(evals): add Vue MFA step-up eval

sanchitmehtagit:
## Summary

- Adds `vue_mfa` eval under `src/evals/mfa/vue/` mirroring the existing `react_mfa` eval
- Scaffold pre-wires `@auth0/auth0-vue` with `createAuth0` and a `TransferView` stub for the agent to add MFA step-up logic
- Graders verified against the [auth0-vue SDK source](https://github.com/auth0/auth0-vue): uses `idTokenClaims` reactive ref (not `getIdTokenClaims()` which isn't exposed by `useAuth0()`)

## Test plan

- [ ] `npm run build` passes
- [ ] `npm test` passes
- [ ] Run `npm run evals -- --eval vue_mfa --mode agent` to smoke-test


yogeshchoudhary147 (changes requested):
The scaffold should be kept generic and minimal. Adding `/transfer` route and `TransferView` stub makes it too specific to this eval. Better to first run a `vue_quickstart` eval and use that output as the `vue/auth0` scaffold, similar to how we did it for React and Angular.

sanchitmehtagit:
Done in 146ee7d — removed the `/transfer` route and `TransferView` stub from the vue/auth0 scaffold; the scaffold is now generic (HomeView only), matching the react/auth0 pattern.

yogeshchoudhary147 (approved):

PR #96, opened 2026-07-09
chore(lint): lint every workspace, not just the app

frederikprijck:
## Summary

Only `apps/auth0-evals` had an ESLint config and a `lint` script, so:

- **`turbo run lint` silently skipped every `packages/*`** — none of the framework packages were ever linted.
- **The `lint-staged` pre-commit hook errored on any `packages/**` file** — it runs bare `eslint --fix` from the repo root, but ESLint 10 flat config resolves the nearest `eslint.config.*` searching up from each file, and no config existed above `packages/*`. Committing framework code required `--no-verify`.

This PR makes linting cover the whole monorepo.

## What's included

- **`eslint.config.base.mjs`** at the repo root — the shared flat config (previously inline in the app).
- **A thin `eslint.config.mjs` in each workspace** that re-exports the base, so ESLint's per-file config resolution finds a config next to any source file (this is what fixes the pre-commit hook) and `turbo run lint` picks each package up.
- **A `lint` script (`eslint .`) in every package**, plus the app's script standardized to `eslint .`.
- **Scope kept to TypeScript sources.** Scaffold templates, generated JS, and `*.config.js` are intentionally excluded — they aren't part of the framework's typechecked source and contain browser/node globals by design. This matches the app's previous `eslint "./**/*.ts"` scope.
- **Fixes for the real violations** this surfaced in the previously-unlinted packages:
  - `eval-core/src/config/loader.ts` — unused `message`; now included in the thrown error (`Failed to load config: …`).
  - `eval/src/cli/config.ts` — removed unused `logger` import.
  - `eval/src/cli/bin.ts` — `eslint-disable no-console` on the top-level CLI error handler (legitimate console use).
  - `eval/src/sandbox/docker.ts` — removed a stale `eslint-disable no-constant-condition` directive.
  - `eval-reporter/src/report.ts` — removed unused `loadScores` from an import (still separately re-exported).
  - `eval-core` tests — replaced two `as any` casts with `as unknown as <Type>`.

## Test plan

- [x] `turbo run lint` passes across all 5 workspaces
- [x] `npm run build` passes
- [x] `npm test` passes (full suite)
- [x] Pre-commit hook (`lint-staged`) now succeeds on `packages/**` files without `--no-verify`

## Summary by CodeRabbit

* **Chores**
  * Standardized linting across the workspace with a shared configuration.
  * Added a `lint` command to multiple packages so checks can be run consistently with `eslint .`.

* **Bug Fixes**
  * Improved config load errors to include the underlying failure message, making setup issues easier to diagnose.
  * Tightened test typings and minor lint-related updates for cleaner, more reliable checks.


frederikprijck:
Thanks @coderabbitai — both addressed in e625a1a:

1. **Scaffold templates now excluded.** Added `**/scaffold/**` and `**/scaffolds/**` to the base config's `ignores`. There were 14 TypeScript scaffold files still being linted; they're seed projects the agent edits (intentional browser/node globals + starter patterns), not framework source. Eval `graders.ts` files live outside those dirs and remain linted.

2. **Config-load error path now tested.** Strengthened the existing broken-config test in `loader.test.ts` to assert the wrapped `EvalConfigError` message includes both the underlying import-failure detail (`Failed to load config:`) and the config path — covering the `loader.ts` message change, not just the error type.

sanchitmehtagit (approved):

PR #108, opened 2026-07-10
fix(reporter): preserve caught error as cause in loadScores

frederikprijck:
## What

Fixes the `preserve-caught-error` lint error currently failing on `main` in `packages/eval-reporter/src/report/processors.ts`.

The `loadScores` catch block rethrows a new `Error` with a contextual message but dropped the original parse error. The `preserve-caught-error` rule (enabled via `eslint.configs.recommended`) flags this. The fix attaches the caught error via the `Error` `cause` option, retaining the underlying stack.

## Why

`npm run lint` fails on `main` today, blocking pre-commit hooks on any branch that merges in latest main.

## Testing

- Added a test asserting `err.cause` is preserved on parse failure.
- `npm run lint` passes across all 5 workspaces.
- `npm test` passes.


yogeshchoudhary147 (approved):
LGTM

PR #165, opened 2026-07-30
feat: configurable proxy auth header

frederikprijck:
## What

Adds an optional, configurable HTTP auth header for the LLM proxy, so the framework can authenticate against a proxy that requires its own header (e.g. LiteLLM's `x-litellm-api-key: Bearer <jwt>`) instead of only the provider-native API-key env vars.

```javascript
// eval.config.js
proxy: {
  baseUrl: PROXY_BASE_URL,
  authHeader: { name: 'x-litellm-api-key', valuePrefix: 'Bearer ', tokenEnv: 'LLM_PROXY_TOKEN' },
}
```

Entirely opt-in. Omit `authHeader` and behaviour is unchanged.

## Two credentials, two switches

`LLM_API_KEY` **takes precedence**, so existing deployments keep working untouched and adopting the header is an explicit act — unset `LLM_API_KEY`.

| `LLM_API_KEY` | `proxy.authHeader` | What authenticates |
| --- | --- | --- |
| set | unset | provider-native API key (default) |
| set | configured | provider-native API key — header ignored (logged) |
| unset | configured | the configured header, token from `tokenEnv` |
| unset | unset | nothing — CLI exits with an error |

Because the header path needs `LLM_API_KEY` absent, `validateApiKey()` no longer hard-exits when it is missing *and* `authHeader` is configured — otherwise the feature would be unreachable.

## Where the header is sent

All seven proxy call sites resolve it from one place (`resolveProxyAuthHeader()`), so the name/prefix/token rules live in exactly one module:

| Site | Mechanism |
| --- | --- |
| claude-code | `ANTHROPIC_CUSTOM_HEADERS` env var |
| codex | `[model_providers.llmproxy.env_http_headers]` referencing a fixed env var name, so the token is never written to `config.toml` |
| gemini-cli | `GEMINI_CLI_CUSTOM_HEADERS` env var (gemini-cli ≥ 0.51) |
| copilot | `ProviderConfig.headers` |
| baseline | `createOpenAI({ headers })` |
| LLM judge | `fetch` headers — drops `Authorization` rather than sending a placeholder |
| recommendation generator | same as the judge |

When the header is configured, provider-native key vars get the inert placeholder `unused-see-proxy-auth-header`. That is load-bearing, not cosmetic: the Gemini CLI's `validateAuthMethod` rejects an empty `GEMINI_API_KEY`, and the Claude binary requires one of its credential vars to be set.

**The token is never logged and never written to disk.** Log lines carry the header *name* only; warnings name the env var, never its contents.

## Verified against a live LiteLLM proxy

| Path | Result |
| --- | --- |
| baseline | grade A |
| LLM judge | 12/13 graders |
| claude-code | grade A (92) |
| codex | grade A (95) |
| gemini-cli | grade A (94) |
| `sandbox.passthroughEnv` → Docker | token forwarded |

Every per-runner mechanism was grounded in the installed SDK source rather than documentation.

## Notes for reviewers

- **Sandbox:** the token's env var is app-named, so it must be added to `sandbox.passthroughEnv`. Omitting it makes sandboxed runs fail while host runs succeed.
- `AGENTS.md` gains a warning that the gemini-cli loopback proxy (`127.0.0.1:9876`) is **not** an obsolete header-injecting shim — it is an SSE-unwrapping workaround for a LiteLLM v1.86.0+ bug and is still required. Bypassing it makes Gemini runs fail with `fetch failed sending request`.
- 1295 tests pass; lint and format clean.

## Summary by CodeRabbit

- **New Features**
  - Added configurable proxy authentication with custom headers, token environment variables, and optional prefixes.
  - Added support across LLM judges, recommendations, baseline evaluations, and supported CLI runners.
  - Added secure credential forwarding for sandboxed runs and provider-specific CLI guidance.
  - Added new grading options, command-order checks, and trace-aware judging support.

- **Bug Fixes**
  - Improved authentication validation and clarified errors when required proxy tokens are missing.

- **Documentation**
  - Updated configuration, architecture, authentication, grading, testing, and contributor guidance.

- **Tests**
  - Added coverage for authentication resolution, runner behavior, fallback handling, precedence, and secret protection.


frederikprijck:
@coderabbitai review

apps/auth0-evals/.env.example
```diff
@@ -24,3 +24,11 @@ CODEX_PROXY_BASE_URL=https://your-llm-proxy.example.com
...
 # For local, when using bedrock, use:
 # CLAUDE_PROXY_BASE_URL=https://your-llm-proxy.example.com/anthropic
+
+# Optional: authenticate against a proxy that requires its own header instead of
+# the provider-native API key. Set BOTH vars below to enable it, and leave
+# LLM_API_KEY unset — that key takes precedence and the header is ignored while
+# it is present. LLM_PROXY_TOKEN is already listed in `sandbox.passthroughEnv`,
+# so it reaches sandboxed runs too.
+# LLM_PROXY_AUTH_HEADER=x-litellm-api-key
+# LLM_PROXY_TOKEN=
```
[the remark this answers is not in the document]
frederikprijck: Fixed in e798b96467186cdac56847fbbb5562e11c2b413a

apps/auth0-evals/eval.config.js
```diff
@@ -6,6 +6,12 @@ const remoteSkillsBranch = process.env.REMOTE_SKILLS_BRANCH || 'main';
 // required at runtime (the CLI fails fast if proxy.baseUrl is empty).
 const PROXY_BASE_URL = process.env.LLM_PROXY_BASE_URL ?? '';
 
+// Name of the HTTP header the proxy authenticates on (e.g. `x-litellm-api-key`).
+// Unset by default, which leaves the framework on its provider-native API-key
+// path. When set, the token in LLM_PROXY_TOKEN rides this header instead —
+// though LLM_API_KEY still takes precedence whenever it is exported.
+const PROXY_AUTH_HEADER_NAME = process.env.LLM_PROXY_AUTH_HEADER ?? '';
+
```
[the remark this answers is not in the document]
frederikprijck: Fixed in e798b96467186cdac56847fbbb5562e11c2b413a

packages/evals/src/cli/validators.ts
```diff
@@ -21,10 +21,32 @@ import {
...
+ * Exits if missing, unless `proxy.authHeader` is configured — that deployment
+ * authenticates via the custom header instead, so requiring the provider-native
+ * key would make the header path unreachable. Returns an empty string in that
+ * case; call sites route the credential through the header and write
+ * `PLACEHOLDER_API_KEY` into provider-native key fields.
+ */
 export function validateApiKey(): string {
   const apiKey = process.env[LLM_API_KEY_ENV];
   if (!apiKey) {
+    if (hasProxyAuthHeader()) return '';
```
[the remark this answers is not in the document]
frederikprijck: Fixed in e798b96467186cdac56847fbbb5562e11c2b413a

sanchitmehtagit:
@coderabbit.ai  generate sequence diagram

packages/evals/src/cli/validators.ts
sanchitmehtagit: Returning `''` here to signal proxy-header mode collides with the judge executor downstream. In `packages/evals-core/src/graders/executors/llm-judge.ts:47` the guard is `if (!ctx.apiKey || !ctx.judge)`, and `''` is falsy, so every judge grader fails with "requires apiKey" before `llmJudge()` runs, even though `llmJudge` (llm-judge.ts:84-87) is written to drop `Authorization` and use the proxy header when the key is empty.

Traced end-to-end: `validateApiKey()` -> `run.ts` -> `runGraders(apiKey='')` -> `ctx.apiKey = ''` -> executor bails. Net effect in authHeader-only mode (LLM_API_KEY unset): the judge/correctness dimension zeroes out. That executor file isn't in this diff, so the new empty-string convention silently trips a pre-existing gate, and no test exercises the executor with `''`.

I'd tighten the executor guard to distinguish "not provided" from "intentionally empty":

```typescript
if (ctx.apiKey === undefined || !ctx.judge) {
```

and add an executor-level test with `apiKey: ''` plus a configured header. Worth re-running the live judge verification with LLM_API_KEY unset, since the "12/13 graders" run likely still had it set.

packages/evals-core/src/config/proxy-auth.ts
```diff
@@ -0,0 +1,91 @@
...
+  // custom header is an explicit act: unset LLM_API_KEY.
+  if (process.env[LLM_API_KEY_ENV]) {
+    logger.info(
+      `[proxy-auth] ${LLM_API_KEY_ENV} is set — using the provider-native API key and ignoring proxy.authHeader.`,
+    );
+    return undefined;
+  }
+
+  const token = process.env[authHeader.tokenEnv];
+  if (!token) {
```
sanchitmehtagit, line 80: When `authHeader` is configured but the token env var is empty, this warns and falls back to `undefined`, so the run proceeds with no working credential and dies later on an opaque 401. I get the rationale in the comment above (a typo shouldn't break every run), but once someone has configured `authHeader` *and* unset `LLM_API_KEY`, both switches are thrown for the header path and an empty token is unambiguously a misconfiguration a loud startup failure would catch faster than a mid-run 401.

Would it make sense to fail closed specifically in that quadrant?

```typescript
if (!token) {
  if (!process.env[LLM_API_KEY_ENV]) {
    logger.error(`[proxy-auth] proxy.authHeader configured but ${authHeader.tokenEnv} is unset and LLM_API_KEY is absent — no credential available.`);
    process.exit(1);
  }
  logger.warn(/* existing fallback message, now only for the LLM_API_KEY-still-set case */);
  return undefined;
}
```

That keeps the soft fallback for the case where a real credential still exists.

sanchitmehtagit (commented):
Solid, well-documented feature: single resolution point, clean secret handling, and thorough docs/tests. Two things worth a look before merge, one of which breaks judge graders in exactly the mode this PR adds.

packages/evals/src/cli/validators.ts
frederikprijck: Fixed in https://github.com/auth0/auth0-evals/pull/165/commits/edb5162867354b811aee6a1bfebd5b3ebe9b7c53

packages/evals-core/src/config/proxy-auth.ts
frederikprijck: Fixed in https://github.com/auth0/auth0-evals/pull/165/commits/edb5162867354b811aee6a1bfebd5b3ebe9b7c53

sanchitmehtagit (dismissed):

PR #178, opened 2026-08-07
feat(evals): add auth0 CLI B2B organization setup eval

developerkunal:
Adds an `auth0_cli_b2b_org_setup` eval under a new `cli` category that measures how well agents configure a complete B2B (organizations) setup using only the `auth0` CLI, against a tenant the CLI is already logged into. It complements the SDK-integration quickstarts with a pure tenant-configuration task.

The prompt is goal-only and names no command or flag, so the agent has to work out the CLI surface itself. It asks for a protected API with `read:reports`/`write:reports`/`manage:members`, two RBAC roles wired to those scopes, a Regular Web App and a Machine-to-Machine app, an M2M client grant between them, two organizations each with an enabled login connection, and an org-admin invitation, then a `smoke-b2b-manifest.json` capturing every created resource.

Grading combines event graders over the agent's successful command trace (`ranCommand`/`ranCommandOneOf`/`wroteFile`) with a trace-aware judge (`includeCommandTrace: true`) that reads the manifest and cross-checks it against the commands actually run. The client grant and org enabled-connections have no public CLI subcommand, so those steps are matched by their Management API endpoint (`client-grants`, `enabled_connections`) rather than a dedicated command. A dedicated L3 judge fails the run if any client secret leaks into the manifest or the trace. Structure mirrors the existing evals: ascending L3/L4/L5 graders and a final levelless holistic judge.

The eval also ships an optional `harness.json` and `verify.js` for a runner that provisions a live Auth0 tenant. Both are inert to the framework here (the loader reads only `PROMPT.md` and `graders.ts`, and tsc compiles only `src/**/*.ts`), so they do not affect build, lint, or discovery.

## Test plan
- [x] `tsc --noEmit` passes for the auth0-evals app
- [x] `eslint` passes on the new `graders.ts`
- [x] `prettier --check` clean on all new files
- [x] Eval discovery finds `auth0_cli_b2b_org_setup` under category `cli`


apps/auth0-evals/src/evals/cli/auth0-cli-b2b-org-setup/verify.js
```diff
@@ -0,0 +1,163 @@
...
+  try {
+    return JSON.parse(r.stdout);
+  } catch {
+    throw new Error(`auth0 api get ${path} did not return JSON: ${r.stdout.slice(0, 200)}`);
+  }
+}
+
+// Management API list endpoints answer either a bare array or a wrapped object
+// (e.g. { clients: [...] }) depending on pagination params. Normalize both.
+function asList(res, key) {
```
sanchitmehtagit, line 50: Nit / totally optional: `asList` is already nicely defensive, and since `api()` throws on both a non-zero exit and non-JSON output, it only ever sees valid parsed JSON here — so this is safe as-is. If you ever want to make a malformed response easier to debug, distinguishing "key present but not an array" from "resource genuinely empty" could help, but definitely not worth holding the PR for. Just flagging in case it's useful later. 🙂

sanchitmehtagit (commented):
Nice eval! The live-tenant `verify.js` is a great addition — asserting the full B2B graph exists on the server (not just in the agent's manifest) closes a gap that the trace and file/judge graders can't cover on their own. The PROMPT/graders structure looks clean and follows the level conventions.

Just one small nit inline, and one non-blocking follow-up that's out of scope for this PR:

- The `ranCommandsInOrder` primitive this eval relies on (already on `main`) passes vacuously if it's ever called with an empty `steps` array — worth a construction-time guard + test at some point, but nothing to change here.

apps/auth0-evals/src/evals/cli/auth0-cli-b2b-org-setup/harness.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$comment": "Optional live-environment descriptor for this eval. The auth0-evals framework never reads this file — its loader only parses PROMPT.md and graders.ts, and tsc compiles src/**/*.ts, so this JSON and the sibling verify.js are inert to discovery and the build. It is a hint for a runner that can provision a live Auth0 tenant: `environment: auth0-tenant` requests a provisioned tenant, `options.escalate` asks the runner to grant the tenant's management client the Management API scopes this flow exercises, and `verify` names the ground-truth script to run against the live tenant before teardown. An eval that needs no live environment simply omits this file.",
+  "environment": "auth0-tenant",
```
sanchitmehtagit: This could be moved to frontmatter for PROMPT.md

developerkunal: Thanks for the suggestion i moved it.

PR #179, opened 2026-08-10
feat(axis): add @a0/evals-axis package

yogeshchoudhary147:
## Summary

- Adds the new `@a0/evals-axis` bridge package that connects AXIS multi-agent orchestration to the existing auth0-evals grader engine
- Exports `runAxis()` and `buildAxisScores()` which are consumed by the app integration (follow-up PR)
- Full test coverage across adapter, grader-hook, report, and run modules

## What this package does

**`adapter.ts`** — Converts AXIS `TranscriptEntry[]` to `EventToolCall[]` for the grader engine. Handles both standard transcript format (Codex, Gemini) and the claude-code nested-assistant format where tool calls are embedded inside assistant entries.

**`grader-hook.ts`** — Plugs into AXIS `onJobComplete`. Discovers the matching eval config, runs compile/setup commands against the agent workspace, and executes L1-L4 graders.

**`report.ts`** — Converts AXIS `ScoredOutput` + `GraderResult[]` into `AgentJobResult[]` (same format as `scores-agent.json`). Correctly derives `tool_calls` via `transcriptAnalysis.toolUseCount`, `active_time` from `sparseIndex`, `interruptions`, `judge_cost_usd`, and per-grader `cost_usd`.

**`run.ts`** — Orchestrates the full AXIS run: starts the run, registers the grader hook, collects results, scores via the judge, and returns structured output.

## Summary by CodeRabbit

* **New Features**
  * Added AXIS evaluation support, including agent execution, scoring, grading, and result reporting.
  * Added transcript conversion for tool calls, with support for errors and structured results.
  * Added score summaries with letter grades, dimensions, costs, tokens, timing, interruptions, and tool-call metrics.
  * Added configurable grader execution, callbacks, debugging, paths, and judge model options.

* **Tests**
  * Added comprehensive automated coverage for transcript conversion, grading, execution, scoring, and reporting.

### DIFF
--- a/packages/evals-axis/src/adapter.ts
+++ b/packages/evals-axis/src/adapter.ts
@@ -1,0 +1,58 @@
+/**
+ * AXIS transcript adapter.
+ *
+ * Converts an AXIS TranscriptEntry[] to the EventToolCall[] format the
+ * auth0-evals grader engine expects.
+ *
+ * AXIS represents tool interactions as paired entries:
+ *   - tool_use:    carries the tool name and input arguments
+ *   - tool_result: carries the output and error flag
+ *
+ * We join pairs by tool ID and produce one EventToolCall per pair.
+ */
+
+import type { TranscriptEntry } from '@netlify/axis';
+import type { EventToolCall } from '@a0/evals-graders';
+
+export function axisTranscriptToToolCalls(transcript: TranscriptEntry[]): EventToolCall[] {
+  // First pass: index tool_use entries by their ID.
+  const toolUseById = new Map<string, { name: string; args: Record<string, unknown> }>();
+
+  for (const entry of transcript) {
+    if (entry.type !== 'tool_use') continue;
+    const c = entry.content as Record<string, unknown>;
+    const id = c['id'] as string | undefined;
+    const name = c['name'] as string | undefined;
+    if (!id || !name) continue;
+    toolUseById.set(id, {
+      name,
+      args: (c['input'] as Record<string, unknown> | undefined) ?? {},
+    });
+  }
+
+  // Second pass: emit EventToolCall for each tool_result paired with a tool_use.
+  const toolCalls: EventToolCall[] = [];
+
+  for (const entry of transcript) {
+    if (entry.type !== 'tool_result') continue;
+    const c = entry.content as Record<string, unknown>;
+    const toolUseId = c['tool_use_id'] as string | undefined;
+    if (!toolUseId) continue;
+
+    const toolUse = toolUseById.get(toolUseId);
+    if (!toolUse) continue;
+
+    const isError = (c['is_error'] as boolean | undefined) ?? false;
+    const contentVal = c['content'];
+    const result = typeof contentVal === 'string' ? contentVal : contentVal != null ? JSON.stringify(contentVal) : '';
+
+    toolCalls.push({
+      name: toolUse.name,
+      args: toolUse.args,
+      result,
+      causedError: isError,
+    });
+  }
+
+  return toolCalls;
+}
--- a/packages/evals-axis/src/grader-hook.ts
+++ b/packages/evals-axis/src/grader-hook.ts
@@ -1,0 +1,70 @@
+/**
+ * AXIS post-run grader hook.
+ *
+ * After an AXIS job completes (and before workspace teardown), this hook
+ * loads the matching eval's graders, optionally runs the compile command,
+ * and calls the auth0-evals grader engine against the workspace.
+ */
+
+import { discoverEvals, loadEval, runGraders, runCompileCommand, runSetupCommand, AGENT_LEVELS } from '@a0/evals-core';
+import type { GraderResult } from '@a0/evals-graders';
+import type { RunResult } from '@netlify/axis';
+import { axisTranscriptToToolCalls } from './adapter.js';
+
+export interface RunAuth0GradersOptions {
+  /** Absolute path to the framework root (contains eval.config.js and src/evals/). */
+  frameworkRoot: string;
+  /** Relative path to the evals directory inside frameworkRoot. Defaults to 'src/evals'. */
+  evalsDir?: string;
+  /** LLM API key forwarded to the judge model. */
+  apiKey: string;
+  /** Judge model identifier. Falls back to the framework config judge model when omitted. */
+  judgeModel?: string;
+}
+
+/**
+ * Runs auth0-evals graders against a completed AXIS job.
+ *
+ * Discovers the eval matching result.scenarioKey, converts the AXIS transcript
+ * to EventToolCall[], runs the compile command (and setup command, if declared)
+ * or the setup command alone when there is no compile step, then grades the
+ * workspace with L1-L4 graders (AGENT_LEVELS).
+ */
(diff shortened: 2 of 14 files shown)


packages/evals-axis/src/grader-hook.ts
```diff
@@ -0,0 +1,70 @@
...
+    );
+  }
+
+  const evalConfigs = discoverEvals(evalsDir, frameworkRoot);
+  const evalConfig = evalConfigs.find((c) => c.id === result.scenarioKey);
+  if (!evalConfig) {
+    throw new Error(`No eval found with id '${result.scenarioKey}' in ${frameworkRoot}/${evalsDir}`);
+  }
+
+  const evalDef = await loadEval(evalConfig, frameworkRoot);
```
sanchitmehtagit, line 48: Test gap: there's no coverage for `loadEval()` **rejecting** (e.g. a corrupt `PROMPT.md` / `graders.ts`). Right now only the resolved path is tested. A rejection here fails grading for the whole job — worth a test asserting the error message names which eval failed so it's debuggable.

packages/evals-axis/src/adapter.ts
```diff
@@ -0,0 +1,58 @@
...
+  const toolCalls: EventToolCall[] = [];
+
+  for (const entry of transcript) {
+    if (entry.type !== 'tool_result') continue;
+    const c = entry.content as Record<string, unknown>;
+    const toolUseId = c['tool_use_id'] as string | undefined;
+    if (!toolUseId) continue;
+
+    const toolUse = toolUseById.get(toolUseId);
+    if (!toolUse) continue;
```
sanchitmehtagit, line 43: This orphaned-`tool_result` skip (a `tool_use_id` that was never indexed) is the right behavior, but there's no test asserting the phantom call is absent from the output. Cheap to add and it locks in the contract. Same goes for the `!id || !name` skip on line 26.

Optional: if AXIS ever changes its transcript shape, these silent `continue`s would drop tool calls with no trace. A debug-level log would make that diagnosable — though the adapter has no logger wired in today, so only worth it if you expect shape churn.

packages/evals-axis/src/run.ts
```diff
@@ -0,0 +1,128 @@
...
+    ...(configPath ? { configPath } : {}),
+    ...(debug ? { debug } : {}),
+    onResult: async (result) => {
+      const key = `${result.scenarioKey}|${result.agentName}`;
+      const reporter = onGraderResults ?? logGraderResults;
+      try {
+        const graderResults = await runAuth0Graders(result, graderOptions);
+        allGraderResults.set(key, graderResults);
+        reporter(result.scenarioKey, result.agentName, graderResults);
+      } catch (err) {
```
sanchitmehtagit, line 65: Minor: this `catch` covers both `runAuth0Graders()` and the `reporter()` call, and then calls `reporter()` *again* on line 76. If a caller passes an `onGraderResults` that throws, the second call throws too and escapes `onResult`. The default `logGraderResults` won't throw so it's fine today — but if you want to be defensive against user-supplied reporters, wrap the in-catch `reporter()` call separately.

packages/evals-axis/src/report.ts
```diff
@@ -0,0 +1,200 @@
...
+    const graders = allGraderResults.get(key) ?? [];
+    const passed = graders.filter((g) => g.passed).length;
+    const judgeCost = computeJudgeCost(graders);
+
+    // agentName format: "claude-code|global.anthropic.claude-opus-4-8"
+    // This pipe-separated convention is set by axis.config.ts agents[] and is
+    // not guaranteed by the AXIS API — if the format changes, rawAgentType
+    // falls through to the claude-code default and model falls back to agentName.
+    const [rawAgentType, model] = result.agentName.split('|');
+    const agentType: AgentType = AXIS_AGENT_TO_TYPE[rawAgentType ?? ''] ?? 'claude-code';
```
sanchitmehtagit, line 108: Small type-safety win: declaring `AXIS_AGENT_TO_TYPE` as `... as const satisfies Record<string, AgentType>` would catch key/value typos at compile time. Also — the comment near the `split('|')` above warns the pipe format "is not guaranteed by the AXIS API," but there's no fallback if the pipe is absent (`model` becomes `undefined`). Either drop the caveat or handle the no-pipe case.

packages/evals-axis/src/report.ts
```diff
@@ -0,0 +1,200 @@
...
+/**
+ * Sums token usage across judge grader results and estimates cost.
+ * Mirrors the private computeJudgeCost helper in @a0/evals-core serializers.
+ */
+function computeJudgeCost(graderResults: GraderResult[]): number {
+  let total = 0;
+  for (const gr of graderResults) {
+    const input = gr.inputTokens ?? 0;
+    const output = gr.outputTokens ?? 0;
+    if ((input > 0 || output > 0) && gr.judgeModel) {
```
sanchitmehtagit, line 42: Edge case worth a decision: a judge grader with `judgeModel` set but 0 input/output tokens (e.g. a cached response) is skipped here and contributes 0 cost. If that's intended, a one-line comment saying "0 tokens = free/cached" would make it explicit; if not, it silently under-reports cost.

sanchitmehtagit (commented):
Overall looks good. some suggestions

packages/evals-axis/src/grader-hook.ts
yogeshchoudhary147: `loadEval` errors bubble up and get caught in the `run.ts` catch block as a `grading-error` result, full error message is preserved there. Happy to add wrapping if this becomes a debugging pain point, but feels premature for now.

packages/evals-axis/src/adapter.ts
yogeshchoudhary147: Both cases are already covered. 'skips tool_result entries with no matching tool_use' covers the orphaned result and 'skips tool_use entries missing id or name' covers the `\!id || \!name` guard, both asserting `toEqual([])`. The `\!name` case isn't a separate fixture but it hits the exact same code path as `\!id` so the contract is locked in. Debug logging is noted but skipping for now as there is no logger wired in the adapter today.

packages/evals-axis/src/report.ts
yogeshchoudhary147: Zero-token case only happens when the judge was never called, not on a cache hit. Even a fully cached response has non-zero output tokens, so the guard is correct. 0 tokens = judge skipped = 0 cost is right. Also intentionally mirrors the same guard in evals-core's `computeJudgeCost`.

packages/evals-axis/src/run.ts
yogeshchoudhary147: Default reporter is just `console.log` so it won't throw. If someone passes a custom `onGraderResults` that throws, that's a bug in their code, not something we need to guard against here.

packages/evals-axis/src/report.ts
yogeshchoudhary147: `as const satisfies` is a style preference with no behaviour change, skipping for now.

On the no-pipe case, it is already handled. Line 162 has `model: model ?? result.agentName` so if there is no `|` the model falls back to the full agentName. The comment is just documenting the convention, not indicating missing handling.

sanchitmehtagit (approved):

PR #196, opened 2026-08-12
feat(evals): add Swift and Android MFA step-up evals

subhankarmaiti:
## ✏️ Changes

Adds MFA step-up evals for the two mobile SDKs: `swift_mfa` and `android_mfa`. Same shape as `mfa/react` — gate a Transfer Funds action behind MFA by requesting step-up through Universal Login with `acr_values` and `max_age` 0, reading `amr` off the ID token, and only then running the transfer. Mobile had no MFA coverage before this.

Both use a new shared scaffold (`src/evals/scaffolds/{swift,android}/auth0`): a login-only app with Auth0 already wired, no Transfer Funds UI, since building the feature is the task. No `compile_command` — there is no native toolchain in the sandbox, same as the mobile quickstarts.

The graders accept either legitimate way of doing each step, so correct solutions aren't failed on style: `amr` via JWTDecode / `customClaims` on Swift and `getExtraInfo()` / jwtdecode on Android, and `max_age` via the dedicated builder or a `max_age` parameter. L2 penalises the embedded MFA grant, which is wrong for a Universal Login app.

Two side fixes: the README eval table listed quickstarts only, so the existing MFA and DPoP evals are backfilled; and `.build/` / `.gradle/` are gitignored, since both scaffolds pick up build residue locally.

## 🔮 Type of Change

- [x] Standard
- [ ] Emergency
- [ ] Significant

## 🔗 References

- SDK-10701, SDK-10702 — under epic SDK-10349
- Reference eval: `apps/auth0-evals/src/evals/mfa/react/`

- [x] I added at least one link (task, Slack thread, etc.) to explain why this change is needed.

## 📖 Documentation

Updated the eval table in `apps/auth0-evals/README.md`. No new primitives, levels or flags, so the other docs stay accurate.

## 🎯 Testing

`npm run build && npm test && npm run lint` pass. No framework logic changed, so no new package tests.

Ran the deterministic graders against the bare scaffolds and against a hand-written correct solution for each platform: everything positive fails on the scaffold, everything passes on the correct solution. That caught two graders the scaffold itself already satisfied, now scoped to the step-up request.

10/10 jobs green:

| Eval | baseline | agent | agent+skills | agent+mcp | agent+mcp+skills |
| --- | --- | --- | --- | --- | --- |
| `swift_mfa` | 5/11 | 10/13 · 87.2 B | 13/13 · 95.8 A | 7/15 · 71.8 C | 11/15 · 84.5 B |
| `android_mfa` | 7/13 | 13/15 · 78.8 B | 13/15 · 83.7 B | 14/17 · 85.1 B | 16/17 · 97 A |

Failures are specific, not noisy: one run never used `acr_values` at all, two hand-rolled Base64 JWT parsing, one set `withMaxAge(0)` only behind a flag the MFA path passed as `false`, one short-circuited the `amr` check on a session boolean.

Ignore the baseline column — two pre-existing framework bugs make it meaningless. `runBaseline` sets no `maxOutputTokens` and doesn't disable thinking, so on Opus 5 the answer goes to reasoning tokens (`swift_mfa` returned 0 characters). And baseline judges always see an empty corpus, because `gradeText` writes to `llm_response.txt` and the judge excludes `.txt`. Follow-ups to come; neither is caused by this PR.

- [ ] This change has integration, unit, or performance test coverage, or I explained why it does not.

### DIFF
--- a/apps/auth0-evals/src/evals/mfa/android/PROMPT.md
+++ b/apps/auth0-evals/src/evals/mfa/android/PROMPT.md
@@ -1,0 +1,14 @@
+---
+id: android_mfa
+name: Android MFA Step-Up
+scaffold: src/evals/scaffolds/android/auth0
+skills: auth0
+---
+
+## Task
+
+My Android app already has Auth0 login working through Universal Login. I want to add a Transfer Funds action where users must complete MFA before the transfer runs. If they haven't done MFA yet, prompt them for it.
+
+Domain: dev-barkbook.us.auth0.com
+Client ID: barkbook_client_abc123xyz
+Audience: https://api.barkbook.com
--- a/apps/auth0-evals/src/evals/mfa/android/graders.ts
+++ b/apps/auth0-evals/src/evals/mfa/android/graders.ts
@@ -23,10 +23,13 @@ export function defineGraders() {
     notContains('auth0-java', 'No auth0-java (server-side SDK, not for Android)', GraderLevel.L2),
     // Auth0.Android has no Credentials.claims property — claims come from credentials.user.
     notContains('credentials.claims', 'No hallucinated Credentials.claims property', GraderLevel.L2),
+    // Every embedded-MFA path carries the mfa_token, whether or not the
+    // MfaApiClient type is ever named.
     notContains(
-      'mfaClient',
-      'Does not use the embedded MFA grant (MFAClient) — wrong approach for a Universal Login app',
+      'mfaToken',
+      'Does not use the embedded MFA grant (MfaApiClient) — wrong approach for a Universal Login app',
       GraderLevel.L2,
+      { caseSensitive: false },
     ),
     notContains('mfa/challenge', 'Does not call the raw MFA challenge endpoint', GraderLevel.L2),
 
--- a/apps/auth0-evals/src/evals/mfa/android/graders.ts
+++ b/apps/auth0-evals/src/evals/mfa/android/graders.ts
@@ -1,0 +1,86 @@
+import { contains, notContains, notContainsInSource, matches, judge, GraderLevel } from '@a0/evals-graders';
+
+export function defineGraders() {
+  return [
+    // ── L1: Required MFA step-up symbols present ───────────────────────────
+    contains('acr_values', 'Step-up request uses acr_values parameter', GraderLevel.L1),
+    contains('amr', 'AMR claim checked to detect prior MFA completion', GraderLevel.L1),
+    contains(
+      'schemas.openid.net/pape/policies/2007/06/multi-factor',
+      'Uses correct multi-factor acr_values policy URI',
+      GraderLevel.L1,
+    ),
+    contains('withParameters', 'Passes acr_values through WebAuthProvider withParameters', GraderLevel.L1),
+    // The SDK surfaces custom claims through UserProfile.getExtraInfo(); the separate
+    // com.auth0.android:jwtdecode library is an equally valid way to read amr.
+    matches(
+      String.raw`getExtraInfo\(\)|com\.auth0\.android:jwtdecode`,
+      'Reads ID token claims via UserProfile.getExtraInfo() or the jwtdecode library',
+      GraderLevel.L1,
+    ),
+
+    // ── L2: Hallucination / wrong approach ────────────────────────────────
+    notContains('auth0-java', 'No auth0-java (server-side SDK, not for Android)', GraderLevel.L2),
+    // Auth0.Android has no Credentials.claims property — claims come from credentials.user.
+    notContains('credentials.claims', 'No hallucinated Credentials.claims property', GraderLevel.L2),
+    notContains(
+      'mfaClient',
+      'Does not use the embedded MFA grant (MFAClient) — wrong approach for a Universal Login app',
+      GraderLevel.L2,
+    ),
+    notContains('mfa/challenge', 'Does not call the raw MFA challenge endpoint', GraderLevel.L2),
+
+    // ── L3: Security ──────────────────────────────────────────────────────
+    notContainsInSource(
+      'barkbook_client_abc123xyz',
+      'No hardcoded client ID in Kotlin source files (ok in strings.xml)',
+      GraderLevel.L3,
+    ),
+    notContainsInSource(
+      'dev-barkbook.us.auth0.com',
+      'No hardcoded domain in Kotlin source files (ok in strings.xml)',
+      GraderLevel.L3,
+    ),
+    judge(
+      'Does the code let SecureCredentialsManager (or CredentialsManager) handle token storage rather than ' +
+        'persisting Auth0 tokens (access tokens, ID tokens, refresh tokens) by hand in SharedPreferences? ' +
+        'Storing application state such as a pending transfer is acceptable — only manual token storage is a violation.',
+      GraderLevel.L3,
+    ),
+
+    // ── L4: Structural correctness ────────────────────────────────────────
+    judge(
+      'Does the code check the amr claim before executing the transfer action, and only proceed when ' +
+        '"mfa" is present in the amr array? Reading amr via credentials.user.getExtraInfo()["amr"] or via ' +
+        'the com.auth0.android:jwtdecode library are both acceptable.',
+      GraderLevel.L4,
+    ),
+    judge(
+      'When MFA is missing, is step-up performed by launching a new WebAuthProvider.login(...) — with ' +
+        'withScheme and the MFA acr_values — rather than by calling the Authentication API MFA endpoints ' +
+        '(challenge/verify) directly?',
+      GraderLevel.L4,
+    ),
+
+    // ── L5: Current API patterns ──────────────────────────────────────────
+    judge(
+      'Does the step-up request force fresh authentication with a max_age of 0, so a cached session is not ' +
+        'reused? Either the dedicated withMaxAge(0) builder or a "max_age" entry inside withParameters(...) counts.',
+      GraderLevel.L5,
+    ),
+    judge(
+      'Is the step-up request built with the current v2+ builder API — WebAuthProvider.login(account) ' +
+        'with withScheme and withParameters/withMaxAge — rather than by hand-building an /authorize URL ' +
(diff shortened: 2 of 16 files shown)


apps/auth0-evals/src/evals/mfa/android/graders.ts
```diff
@@ -0,0 +1,86 @@
...
+  return [
+    // ── L1: Required MFA step-up symbols present ───────────────────────────
+    contains('acr_values', 'Step-up request uses acr_values parameter', GraderLevel.L1),
+    contains('amr', 'AMR claim checked to detect prior MFA completion', GraderLevel.L1),
+    contains(
+      'schemas.openid.net/pape/policies/2007/06/multi-factor',
+      'Uses correct multi-factor acr_values policy URI',
+      GraderLevel.L1,
+    ),
+    contains('withParameters', 'Passes acr_values through WebAuthProvider withParameters', GraderLevel.L1),
```
sanchitmehtagit, line 13: The `withParameters` check at L1 creates a false-negative risk. Any correct solution that uses only `withMaxAge(0)` (without calling `withParameters`) will fail this grader despite being correct. The `contains('acr_values', ...)` check on line 6 already verifies the parameter is present somewhere — this method-specific check is redundant.

**Suggested fix:** Remove this line. If you want to verify the current builder API is used, that belongs in an L5 judge.

apps/auth0-evals/src/evals/mfa/android/graders.ts
```diff
@@ -0,0 +1,86 @@
...
+      'Reads ID token claims via UserProfile.getExtraInfo() or the jwtdecode library',
+      GraderLevel.L1,
+    ),
+
+    // ── L2: Hallucination / wrong approach ────────────────────────────────
+    notContains('auth0-java', 'No auth0-java (server-side SDK, not for Android)', GraderLevel.L2),
+    // Auth0.Android has no Credentials.claims property — claims come from credentials.user.
+    notContains('credentials.claims', 'No hallucinated Credentials.claims property', GraderLevel.L2),
+    notContains(
+      'mfaClient',
```
sanchitmehtagit: This grader checks for lowercase `'mfaClient'` (variable/method form) but misses the uppercase `'MFAClient'` class name. An agent that imports the class without assigning it to a `mfaClient`-named variable bypasses this check:

```kotlin
import com.auth0.android.provider.MFAClient  // 'mfaClient' never appears → grader passes
val client = MFAClient.create(...)           // same
```

**Suggested fix:** Add a second grader after this one:
```typescript
notContains('MFAClient', 'No hallucinated MFAClient class import', GraderLevel.L2),
```

apps/auth0-evals/src/evals/mfa/swift/graders.ts
```diff
@@ -0,0 +1,85 @@
...
+    // dependency of the SDK) or CredentialsManager.userProfile()?.customClaims.
+    matches(
+      String.raw`decode\(jwt:|customClaims`,
+      'Reads ID token claims via JWTDecode or UserProfile.customClaims',
+      GraderLevel.L1,
+    ),
+
+    // ── L2: Hallucination / wrong approach ────────────────────────────────
+    notContains('Auth0SDK', 'No hallucinated Auth0SDK package name (correct package is Auth0)', GraderLevel.L2),
+    notContains(
```
sanchitmehtagit, line 25: Same issue as the Android grader — checks for lowercase `'mfaClient'` but misses the uppercase `'MFAClient'` class name. If an agent imports or references the class by its proper PascalCase name without using a lowercase-named variable, the hallucination goes undetected.

**Suggested fix:** Add a second grader after this one:
```typescript
notContains('MFAClient', 'No hallucinated MFAClient class', GraderLevel.L2),
```

apps/auth0-evals/src/evals/mfa/swift/graders.ts
```diff
@@ -0,0 +1,85 @@
...
+    judge(
+      'When MFA is missing, is step-up performed by starting a new Universal Login web auth session — ' +
+        'Auth0.webAuth() with the MFA acr_values — rather than by calling the Authentication API MFA ' +
+        'endpoints (challenge/verify) directly?',
+      GraderLevel.L4,
+    ),
+
+    // ── L5: Current API patterns ──────────────────────────────────────────
+    // `.parameters` / `.maxAge` are the current builders for extra authorization
+    // parameters. The scaffold's login call uses neither, so this only passes if
```
sanchitmehtagit, line 67: This deterministic `matches` check is location-agnostic — it only verifies `.parameters(` or `.maxAge(` exists *somewhere* in the workspace, not that it's used in the step-up request specifically. The L4 judge already validates the step-up flow is correct; this check adds little and could fail correct solutions that pass `acr_values` via a different-but-valid mechanism.

**Suggested fix:** Either drop this in favor of relying on the L5 judge below it (line 71), or add a comment making the assumption explicit.

apps/auth0-evals/src/evals/mfa/android/PROMPT.md
```diff
@@ -0,0 +1,14 @@
+---
+id: android_mfa
+name: Android MFA Step-Up
+scaffold: src/evals/scaffolds/android/auth0
+skills: auth0
+---
+
+## Task
+
+My Android app already has Auth0 login working through Universal Login. I want to add a Transfer Funds action where users must complete MFA before the transfer runs. If they haven't done MFA yet, prompt them for it.
```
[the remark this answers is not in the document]
subhankarmaiti: Intentional. The prompt is the user's ask; the graders encode what a correct implementation looks like. `mfa/react` is worded the same way and asserts the same `acr_values` / `max_age` / `amr` criteria — the eval measures whether the agent knows the step-up mechanics from training data, skills or MCP, so putting them in the prompt turns it into a dictation test. Correct solutions do pass as written: 13/13 on `swift_mfa` with skills, 16/17 on `android_mfa` with mcp+skills.

apps/auth0-evals/src/evals/scaffolds/swift/auth0/Package.swift
```diff
@@ -0,0 +1,24 @@
...
+    dependencies: [
+        .package(url: "https://github.com/auth0/Auth0.swift.git", from: "3.0.0"),
+    ],
+    targets: [
+        .target(
+            name: "BarkbookApp",
+            dependencies: [
+                .product(name: "Auth0", package: "Auth0.swift"),
+            ]),
+    ]
```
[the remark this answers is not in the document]
subhankarmaiti: The mismatch is real, but the suggested fix isn't available: SwiftPM can't express an iOS app bundle (no app product, no `Info.plist`, no URL scheme), which is also why `Auth0.plist` sits at the package root rather than as a target resource — it's the realistic non-source home for the client ID, which is what `notContainsInSource` allows. Nothing compiles in this eval: there's no `compile_command`, and the scaffold's `AGENTS.md` tells the agent not to build. Leaving as is.

apps/auth0-evals/src/evals/mfa/android/graders.ts
subhankarmaiti: Checked `WebAuthProvider.kt` on `main`: the login builder has `withMaxAge(Int)` and `withParameters(Map<String, Any?>)`, no singular `withParameter` and no `acr_values` builder. So `acr_values` can only reach the request through `withParameters` — anything that fails this grader also fails `contains('acr_values')`. Near-redundant, agreed, but it catches one thing that grader can't: an agent hand-building an `/authorize` URL with `acr_values`. Keeping it.

apps/auth0-evals/src/evals/mfa/android/graders.ts
subhankarmaiti: Good catch, and it's worse than the casing: `notContains` defaults to `caseSensitive: true`, and the needle names the wrong symbol on each platform. In Auth0.swift `MFAClient` is a real protocol reached via `Auth0.mfa()`, so the type may never appear in the code at all; in Auth0.Android the class is `MfaApiClient`, which `mfaclient` doesn't match even case-insensitively — so this grader couldn't fire on Android.

Switched both to the one thing every embedded-MFA path must carry, case-insensitive:

```typescript
notContains(
  'mfaToken',
  'Does not use the embedded MFA grant (MfaApiClient) — wrong approach for a Universal Login app',
  GraderLevel.L2,
  { caseSensitive: false },
),
```

Went with a replacement rather than a second grader — an L2 slot that can never fail would inflate the Hallucination dimension.

apps/auth0-evals/src/evals/mfa/swift/graders.ts
subhankarmaiti: Good catch, and it's worse than the casing: `notContains` defaults to `caseSensitive: true`, and the needle names the wrong symbol on each platform. In Auth0.swift `MFAClient` is a real protocol reached via `Auth0.mfa()`, so the type may never appear in the code at all; in Auth0.Android the class is `MfaApiClient`, which `mfaclient` doesn't match even case-insensitively — so this grader couldn't fire on Android.

Switched both to the one thing every embedded-MFA path must carry, case-insensitive:

```typescript
notContains(
  'mfaToken',
  'Does not use the embedded MFA grant (Auth0.mfa()/MFAClient) — wrong approach for a Universal Login app',
  GraderLevel.L2,
  { caseSensitive: false },
),
```

Went with a replacement rather than a second grader — an L2 slot that can never fail would inflate the Hallucination dimension.

apps/auth0-evals/src/evals/mfa/swift/graders.ts
subhankarmaiti: Location-agnostic, true, but there's no valid alternative mechanism to miss: Auth0.swift's WebAuth builder has no `acr_values` method, so `.parameters([String: String])` is the only way to pass it, and `.maxAge(Int)` is the only dedicated `max_age` builder. The scaffold's login call uses neither, so this can only pass if the agent built the step-up request. Keeping it as the deterministic counterpart to the judge below it.

sanchitmehtagit (approved):

PR #198, opened 2026-08-13
fix(graders): thread agent reply text into graders so MCP-only evals grade correctly

sanchitmehtagit:
## ✏️ Changes

For hosted-MCP evals where the agent replies with text instead of writing files, `judge()` and `contains()` always silently scored 0. `collectFiles()` returned an empty map, so every text-based grader treated the workspace as blank regardless of what the agent said. The log signature was `[judge] 0 files, 0 chars total` and a verdict of "the code block is empty" on every run, no matter the actual agent output.

The root cause was that `record.finalSummary` (the agent's reply text) was tracked internally but never passed into `runGraders()`, so `GraderContext` had no way to expose it to executors.

**What changed:**

Commit 1 -- thread agent reply text into runGraders:

- Added `agentText?: string` to `GraderContext` (`executors/types.ts`)
- Added `agentText?` as a trailing optional param to `runGraders()` in `engine.ts` and threaded it into the context object
- Both call sites in this repo (`run.ts`, `sandbox-runner.ts`) now pass `record.finalSummary` as `agentText`

Commit 2 -- fix the AXIS call site:

- `packages/evals-axis/src/grader-hook.ts` now passes `result.output.result` as `agentText`, closing the third call site that was previously missing

Commit 3 -- opt-in `source` field (replaces earlier auto-fallback design):

- Added `source?: 'files' | 'response' | 'both'` to `GraderDef` and `GraderOptions` in `@a0/evals-graders`; default is `'files'` -- zero behavior change for every existing eval
- Wired `source` through `contains`, `notContains`, `matches` factory functions and added it to `JudgeOptions` on `judge()`
- Rewrote `contains`, `notContains`, `matches`, and `llm-judge` executors to be opt-in only: agent reply text is only consulted when the grader explicitly declares `source: 'response'` or `source: 'both'` -- no silent fallback
- Tests replaced auto-fallback assertions with explicit `source` cases and added a regression guard confirming the default still ignores `agentText`
- Updated AGENTS.md grader primitives table to document the `source` option

Commit 4 -- docs and baseline threading (separate commits):

- Updated ARCHITECTURE.md sequence diagrams to show `agentText` flow and `source` opt-in
- `gradeText` (baseline mode) now passes the full response text as `agentText` so `source: 'response'` works consistently across both agent and baseline modes

Commit 5 -- fix llm-judge file-collection gating and add judge+source tests:

- Previously `judgeEntries` was built unconditionally from all workspace files regardless of `source`, so a `source: 'response'` judge sent files+reply instead of reply-only -- contradicting the documented contract and consuming `maxCodeChars` budget on scaffold code the judge never needed
- Fixed by hoisting `source`/`includeFiles` before `judgeEntries` and gating: `const judgeEntries = includeFiles ? Object.entries(ctx.files).filter(...) : []`
- Added 4 judge+source tests to `engine.test.ts`: `source:'response'` sends reply only (scaffold absent from corpus), `source:'response'` passes when LLM returns yes, `source:'both'` sends both files and reply, default `source` does not include agent reply (regression guard)

**Design rationale:** The opt-in approach keeps the grader author in control. Auto-fallback (search `agentText` when files don't match) would silently pass a file-based grader for an agent that forgot to write the file but mentioned the needle in its reply -- violating AGENTS.md principle #1.

**How to use in a new MCP eval grader:**

```typescript
// Agent calls MCP tools and replies with text; never writes files.
contains('useAuth0', 'uses useAuth0 hook', GraderLevel.L1, { source: 'response' }),
judge('Did the agent correctly wire Auth0 to the React app?', undefined, { source: 'response' }),
```

All existing file-based evals are unaffected: `agentText` is optional and the default `source` is `'files'`.

## 🔮 Type of Change

- [x] Standard
- [ ] Emergency
- [ ] Significant

## 🔗 References

This was identified while analyzing MCP-only eval runs where event-based graders passed (because tool call args were correct) but every `judge()` and `contains()` grader failed with "0 files, 0 chars total".

## 📖 Documentation

Updated AGENTS.md grader primitives table to document the `source` option on `contains`, `notContains`, `matches`, and `judge`. Updated ARCHITECTURE.md sequence diagrams to show `agentText` flow. No other public API surface changed.

## 🎯 Testing

The `agentText` and `source` paths are covered by 12 unit tests in `engine.test.ts`: 8 covering `contains`/`notContains`/`matches` with all `source` values plus a regression guard, and 4 covering the `judge` executor's `source` paths (including the corpus-content assertion that would have caught the file-collection gating bug). All existing tests continue to pass (523 tests across evals-core).

- [x] This change has integration, unit, or performance test coverage, or I explained why it does not.

## 🚀 Deployment

Library change only -- no service deployment. The `agentText` param is additive and optional; the `source` field defaults to `'files'` so no existing grader needs updating.

## 🔥 Rollback

Revert the PR. MCP-only eval graders return to always scoring 0 on `judge()` and `contains()`, which was the pre-fix silent failure mode.

## References

- [GitHub Actions](https://github.com/auth0/auth0-evals/actions?query=branch%3Afix%2Fagent-text-grader-support)

## Summary by CodeRabbit

* **New Features**
  * Graders can evaluate workspace files, the agent’s final response, or both.
  * Added source selection to text-matching and judge graders, with workspace files remaining the default.
  * Matching results now identify where content was found or searched.
* **Documentation**
  * Updated grader guidance and architecture documentation to explain response-based evaluation and source selection.


sanchitmehtagit (commented):
Good feature overall — the `source` opt-in is clean and the default-to-`'files'` correctly preserves existing behavior. A few correctness issues worth addressing before merge.

packages/evals-core/src/graders/executors/llm-judge.ts
```diff
@@ -67,7 +67,11 @@ export const llmJudgeExecutor: GraderExecutor = {
     const { model, baseUrl, maxTokens, maxCodeChars, enforceMaxChars } = ctx.judge;
 
     const judgeEntries = Object.entries(ctx.files).filter(([k]) => !isJudgeExcluded(k));
-    const judgeText = judgeEntries.map(([k, v]) => `// FILE: ${k}\n${v}`).join('\n\n');
+    let judgeText = judgeEntries.map(([k, v]) => `// FILE: ${k}\n${v}`).join('\n\n');
+    if (ctx.agentText) {
+      const separator = judgeText ? '\n\n' : '';
+      judgeText = `${judgeText}${separator}// AGENT REPLY:\n${ctx.agentText}`;
+    }
```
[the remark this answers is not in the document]
sanchitmehtagit: The PR already addresses this — 4 judge+source tests cover exactly these paths (lines 1460–1574 in `engine.test.ts`): empty workspace + `source:'response'`, `source:'response'` passes when LLM says yes, `source:'both'` sends files and reply, and a regression guard confirming the default omits agent reply.

packages/evals-core/tests/graders/engine.test.ts
```diff
@@ -1315,3 +1315,261 @@ describe('runGraders - compile', () => {
...
+      'unused',
+      undefined,
+      undefined,
+      true,
+      undefined,
+      undefined,
+      'Use a different hook.',
+    );
+    expect(results[0]!.passed).toBe(false);
+  });
```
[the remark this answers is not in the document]
sanchitmehtagit: Fixed in commit 291940b. Added three new `source:'both'` tests that prove agentText is read:\n- `contains`: passes when needle is only in `agentText` (no workspace files)\n- `notContains`: fails when needle is only in `agentText`\n- `matches`: passes when pattern matches only in `agentText`

packages/evals-core/tests/graders/engine.test.ts
```diff
@@ -1315,3 +1315,261 @@ describe('runGraders - compile', () => {
...
+      dir,
+      'key',
+      undefined,
+      undefined,
+      true,
+      undefined,
+      undefined,
+      agentReply,
+    );
+    const messages = capturedBody!.messages as Array<{ role: string; content: string }>;
```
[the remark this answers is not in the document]
sanchitmehtagit: Kept the existing approach. Each `const messages` is declared inside its own `it(...)` closure — they are in separate block scopes, not the same block. TypeScript accepts this without error and all 526 tests pass. This is a false positive.

packages/evals-graders/src/types.ts
```diff
@@ -63,10 +63,26 @@ export interface GraderDef {
...
+  source?: 'files' | 'response' | 'both';
 }
 
 export interface GraderOptions {
   caseSensitive?: boolean;
+  /**
+   * Where to search — see `GraderDef.source` for semantics.
+   * Defaults to `'files'` (workspace files only).
+   */
+  source?: 'files' | 'response' | 'both';
```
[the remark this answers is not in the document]
sanchitmehtagit: Fixed in commit 291940b. Narrowed `notContainsInSource`'s options parameter to `Pick<GraderOptions, 'caseSensitive'>`, so callers can no longer pass a `source` option that would be silently ignored.

AGENTS.md
```diff
@@ -88,6 +88,30 @@ Use `notContainsInSource` (not `notContains`) when a value like a client ID is a
...
+
+```typescript
+contains('useAuth0', 'uses useAuth0 hook', GraderLevel.L1, { source: 'response' }),
+notContains('auth0-js', 'no legacy SDK', GraderLevel.L2, { source: 'response' }),
+judge('Did the agent correctly describe wiring Auth0 to the React app?', undefined, { source: 'response' }),
+```
+
+- For `judge`, `source: 'response'` skips file collection entirely — the judge corpus is the agent's reply only, so scaffold files cannot inflate it past `maxCodeChars`.
+- For `contains` / `notContains` / `matches`, `source: 'both'` is useful when the agent both writes files **and** summarises the result in its reply and you want to check either.
+- All existing evals are unaffected: `source` defaults to `'files'` and `agentText` is `''` in baseline mode.
```
subhankarmaiti: `gradeText` passes the full response text as `agentText` (grade-text.ts:43), so this line and the `agentText` JSDoc are both wrong for baseline. Worth noting too that `source: 'both'` in baseline would then match prose, not the extracted code.

AGENTS.md
```diff
@@ -75,11 +75,11 @@ Use `notContainsInSource` (not `notContains`) when a value like a client ID is a
...
 | ------------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `contains(needle)`                               | Substring present in any non-excluded workspace file                                                                                                                                                                                           |
-| `notContains(needle)`                            | Substring must NOT appear in any non-excluded workspace file                                                                                                                                                                                   |
+| `contains(needle)`                               | Substring present in any non-excluded workspace file. Pass `{ source: 'response' }` or `{ source: 'both' }` in options to also (or exclusively) search the agent's final reply text — use for MCP-only evals that write no files.             |
```
subhankarmaiti, line 78: `docs/ADDING_EVALS.md:109` still says options is just `caseSensitive`, so the `source` option is undocumented there. The doc table in this file lists ADDING_EVALS.md for grader primitive changes.

packages/evals-core/src/graders/executors/search-corpus.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+import type { GraderContext } from './types.js';
+
+export interface SearchResult {
+  inFiles: boolean;
+  inAgent: boolean;
+}
+
+export function searchCorpus(ctx: GraderContext, needle: string, caseSensitive: boolean, source: string): SearchResult {
```
subhankarmaiti: `source` is typed `string` here, but `GraderSource` is exported from `@a0/evals-graders` and every other executor imports its types from there. Tightening it would catch a typo'd source value at compile time.

packages/evals-core/src/graders/executors/matches.ts
```diff
@@ -18,8 +19,20 @@ export const matchesExecutor: GraderExecutor = {
       // Case-insensitive by default; honor caseSensitive to match the other
       // text-search executors (contains/notContains). Always multiline.
       const flags = def.caseSensitive === true ? 'm' : 'im';
-      passed = new RegExp(pattern, flags).test(ctx.combinedText);
-      detail = `/${pattern}/ ${passed ? 'matched' : 'NOT matched'}`;
+      const re = new RegExp(pattern, flags);
+      const source = def.source ?? 'files';
+
+      const checkFiles = source === 'files' || source === 'both';
```
subhankarmaiti: `contains` and `notContains` both go through the new `searchCorpus` helper, but this reimplements the same scope logic inline and duplicates the `searchScope` ternary verbatim from contains.ts:24. Any reason not to reuse it here?

AGENTS.md
```diff
@@ -88,6 +88,30 @@ Use `notContainsInSource` (not `notContains`) when a value like a client ID is a
 | `calledTool(toolName, description, level)` | Agent invoked an MCP tool whose name contains `toolName` (case-insensitive substring against `mcp__<server>__<tool>` calls; errored calls excluded) — event-based, level required (L4 or L5) |
 | `calledToolOneOf(toolNames, description, level)` | Agent invoked at least one of the named MCP tools — event-based, level required (L4 or L5) |
 
+### Grading the agent's final answer (MCP-only evals)
+
+For hosted-MCP evals the agent calls tools and replies with text — it never writes files. Every text-based grader defaults to `source: 'files'`, so without the `source` option the workspace corpus is empty and every grader silently scores 0.
+
```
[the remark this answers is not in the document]
sanchitmehtagit: Fixed in 51506a39. Updated the wording to clarify that text-based graders search workspace files by default, negative checks (`notContains`) can still pass on an empty corpus, and only graders that explicitly set `source: 'response'` or `source: 'both'` will inspect the agent reply.

AGENTS.md
sanchitmehtagit: Fixed in 51506a39. Corrected the doc: `agentText` in baseline mode is the full LLM response text (passed via `gradeText` → `runGraders`), not `''`. Also noted that `source: 'both'` in baseline searches the raw prose reply, not the extracted code written to disk.

AGENTS.md
sanchitmehtagit: Fixed in 51506a39. Expanded the options description in `docs/ADDING_EVALS.md` to a table that documents both `caseSensitive` and the new `source` option, including its valid values and when to use `'response'` / `'both'` for MCP-only evals.

packages/evals-core/src/graders/executors/search-corpus.ts
sanchitmehtagit: Fixed in 51506a39. Changed the `source` parameter type in `searchCorpus` from `string` to `GraderSource` (imported from `@a0/evals-graders`). Typo'd source values now produce a TypeScript error at compile time.

packages/evals-core/src/graders/executors/matches.ts
sanchitmehtagit: Fixed in 51506a39. Added `searchCorpusRegex(ctx, re, source)` to `search-corpus.ts` (parallel to `searchCorpus` but takes a `RegExp`) and updated `matches.ts` to call it, removing the duplicated `checkFiles`/`checkResponse` ternary logic.

subhankarmaiti (approved):

PR #210, opened 2026-08-19
feat(evals): add Express API DPoP, RBAC and multi-custom-domain evals

tanya732:
## ✏️ Changes

Adds three evals covering `express-oauth2-jwt-bearer` surfaces that `express_api_quickstart` doesn't reach: `express_api_dpop`, `express_api_rbac`, `express_api_mcd`.

Each eval targets a real SDK trap that type-checks and runs:

- **`requiredScopes('a', 'b')`** silently drops every argument after the first — the call succeeds and under-enforces. Only the single-string and array forms pass.
- **`mcd`** cannot be combined with `issuerBaseURL`/`issuer`/`jwksUri`, and `issuerBaseURL` defaults silently from `ISSUER_BASE_URL`. The scaffold ships that variable populated, so the agent must actually remove it or `auth()` throws at startup.
- **`dpop.required` defaults to `false`** while `dpop.enabled` defaults to `true` — a `dpop` object without `required: true` accepts plain Bearer tokens. (Note: `{ enabled: false, required: true }` is *not* a silent misconfiguration; it throws at construction.)

Graders are deliberately tight on the trap and permissive on style: hoisted constants, ES6 shorthand, `split(/\s*,\s*/)` and namespaced custom claims all pass, since Auth0 requires Action-added claims to be namespaced. Regex needles carry `(?<![A-Za-z])` guards because `matches` is case-insensitive and unanchored — without them `required:\s*true` also matches `authRequired: true` and `issuers:` also matches the invented `allowedIssuers:`.

**No framework code touched** — no new primitives, levels or flags, so the remaining docs stay accurate.

### Validation

Each eval run three ways. Solutions deliberately use the awkward-but-legal forms (hoisted `mcd`, `{ issuers }` shorthand, namespaced `isAdmin`) so passing exercises the graders rather than the happy path.

| Eval | Bare scaffold | Correct solution | Wrong solution |
| --- | --- | --- | --- |
| `express_api_dpop` | 14/23 | **23/23** | 20/23 — 3 caught |
| `express_api_rbac` | 16/25 | **25/25** | 22/25 — 3 caught |
| `express_api_mcd` | 14/21 | **21/21** | 18/21  — 3 caught |

Wrong solutions caught include two-argument `requiredScopes`, `claimEquals('isAdmin', 'true')` with a string, `claimIncludes('scope', …)` against the wrong claim, and an invented `allowedIssuers` option.

## 🔮 Type of Change

- [x] Standard
- [ ] Emergency
- [ ] Significant


docs/EVAL_AUTHORING_PLAYBOOK.md
```diff
@@ -0,0 +1,761 @@
+# Eval Authoring Playbook
```
sanchitmehtagit: Thanks for adding this . We also have ADDING_EVALS.md may be we can merge the docs

sanchitmehtagit (dismissed):
LGTM

sanchitmehtagit (approved):

PR #211, opened 2026-08-20
feat(evals): add MFA API evals for auth0-auth-js and auth0-server-js

subhankarmaiti:
Adds two evals covering the MFA APIs in `@auth0/auth0-auth-js` and `@auth0/auth0-server-js` — the flow where an app that exchanges credentials for tokens itself gets `mfa_required` back and has to drive enrollment, challenge and verification through the SDK instead of a hosted login page. The existing `mfa/*` evals all cover step-up via `acr_values`, so nothing measured this surface.

Each eval starts from a minimal Express + TypeScript service with no MFA in it and grades the written code. The two diverge where the SDKs do: the lower-level client exposes factor deletion and the server client does not, so one eval requires it and the other checks it is absent. Credentials come from the prompt and stay out of source, as in the other server-side evals.

`@auth0/auth0-api-js` has no MFA surface, so it gets no eval.

### DIFF
--- a/apps/auth0-evals/src/evals/mfa/auth0-auth-js/PROMPT.md
+++ b/apps/auth0-evals/src/evals/mfa/auth0-auth-js/PROMPT.md
@@ -1,0 +1,25 @@
+---
+id: auth0_auth_js_mfa
+name: Auth0 Auth JS MFA
+scaffold: src/evals/scaffolds/auth0-auth-js/auth0
+skills: auth0
+setup_command: npm install
+compile_command: npm run build
+---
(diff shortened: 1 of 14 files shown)


subhankarmaiti:
Superseded by #213, an independent re-derivation. That version fixes three defects here: the storeOptions judge asserted all the MFA methods take store options as a second argument when only verify does, a notContainsInSource check asserted a session-secret value that appears in neither the prompt nor the scaffold, and challenge_type was missing from the server-js L2 set.

PR #213, opened 2026-08-21
feat(evals): add MFA API evals for auth0-auth-js and auth0-server-js

subhankarmaiti:
Two evals covering the MFA API path, where the app exchanges credentials itself and has to drive enrollment, challenge and verify rather than letting Universal Login prompt. Both mirror the existing step-up siblings under `mfa/`.

`auth0_auth_js_mfa` builds HTTP routes for a mobile client: authenticator-app setup with a QR code, challenge for an already-enrolled user, recovery-code sign-in, and listing and removing factors. `auth0_server_js_mfa` does the same for a session-backed web app, plus the SMS path, and requires the session to end up signed in so `/profile` and the transfers call keep working.

Scaffolds are minimal Express + TypeScript, one per package, with no MFA wiring, so the L1 needles can only come from the model's own output.

`@auth0/auth0-api-js` is not covered: it validates tokens and has no MFA surface.

Replaces #211, which had three defects this derivation caught:

- the `storeOptions` judge asserted that all the MFA methods take store options as a second argument; only `verify` does (`server-mfa-client.ts:30,41,52,68`)
- a `notContainsInSource` check on a session-secret value that appears in neither the prompt nor the scaffold, so it always passed
- `challenge_type` was missing from the server-js L2 set while `binding_code` was present

Both now pass every grader on `agent+mcp+skills` with `claude-opus-5` :

| Eval | Graders | Overall | Correctness | Hallucination | Security |
|---|---|---|---|---|---|
| `auth0_auth_js_mfa` | 34/34 | 98.4 A | 100 | 100 | 100 |
| `auth0_server_js_mfa` | 40/40 | 93.7 A | 100 | 100 | 100 |

Getting there took two rounds of grader fixes. The last was the holistic judge in both files, which opened with "judge against the installed package types, not recall". That is unsatisfiable, because `node_modules` is excluded from the grading corpus, so the types were never in the judge's input. The judge fell back on recall anyway, cited the instruction as grounds to reject, and failed code it had just agreed was structurally complete, flagging real fields (`recoveryCodes`, `oobCode`, `bindingCode`) as invented. Both holistic judges now use the three-clause shape the react, vue, angular, swift and android siblings use, naming symbols the judge can find in the corpus.

Still unmeasured: that job runs one configuration and one model, with no baseline, so nothing here shows the graders discriminate. Worth a baseline and a second model before treating them as calibrated.

The server-js Docs Quality 66 is an allowlist gap, not an agent problem. The agent fetched `raw.githubusercontent.com/auth0/auth0-auth-js/main/packages/auth0-server-js/MFA.md` successfully, and that host is not in `scoring.docUrlSources` in `eval.config.js`, so the lookup scored 33 + 33 + 0. Adding it is a separate change. The server-js Setup Friction 90 and Error Recovery 80 are one provider error.

One deliberate deviation from `docs/ADDING_EVALS.md`: no committed `.env.example`, and so no `wroteFile('.env')` grader. `compile_command` is `tsc` and needs no environment.

The `speakeasy` and `otplib` L2 needles have no field evidence. They are carried over because `mfa/react/graders.ts:16-17` already asserts them and the siblings are the specification.

## Summary by CodeRabbit

- **New Features**
  - Added MFA evaluation scenarios covering enrollment, authenticator and SMS challenges, verification, recovery codes, factor management, session completion, and token handling.
  - Added authentication service scaffolds supporting login, token refresh, hosted authentication flows, profile access, logout, and authenticated API requests.
  - Added secure cookie-based session handling and environment-based Auth0 configuration.
- **Evaluation Improvements**
  - Added automated checks for SDK usage, security practices, compilation, API structure, and complete MFA behavior.
- **Chores**
  - Updated the evaluation judge configuration to use the available model version.

### DIFF
--- a/apps/auth0-evals/eval.config.js
+++ b/apps/auth0-evals/eval.config.js
@@ -80,9 +80,8 @@ export default {
   },
 
   judge: {
-    // Opus 4.5 while Opus 5 is unavailable.
-    model: 'claude-sonnet-5',
-    // Headroom so a verbose judge still reaches its final yes/no line. Opus
+    model: 'claude-opus-5',
+    // Headroom so a verbose judge still reaches its final yes/no line. Opus 5
     // counts thinking against max_tokens; the judge disables thinking, but this
     // also covers proxies that ignore that flag.
     maxTokens: 4096,
--- a/apps/auth0-evals/eval.config.js
+++ b/apps/auth0-evals/eval.config.js
@@ -80,8 +80,9 @@ export default {
   },
 
   judge: {
-    model: 'claude-opus-5',
-    // Headroom so a verbose judge still reaches its final yes/no line. Opus 5
+    // Opus 4.5 while Opus 5 is unavailable.
+    model: 'claude-sonnet-5',
+    // Headroom so a verbose judge still reaches its final yes/no line. Opus
     // counts thinking against max_tokens; the judge disables thinking, but this
     // also covers proxies that ignore that flag.
     maxTokens: 4096,
--- a/apps/auth0-evals/src/evals/mfa/auth0-server-js/graders.ts
+++ b/apps/auth0-evals/src/evals/mfa/auth0-server-js/graders.ts
@@ -23,20 +23,15 @@ export function defineGraders() {
       GraderLevel.L2,
     ),
     notContains('.authClient.mfa', 'Does not reach through to the internal auth client', GraderLevel.L2),
-    notContains('mfa/associate', 'Does not hand-roll a call to the /mfa/associate endpoint', GraderLevel.L2),
-    notContains('mfa/challenge', 'Does not hand-roll a call to the /mfa/challenge endpoint', GraderLevel.L2),
     notContains('/oauth/token', 'Does not hand-roll a raw token grant', GraderLevel.L2),
     notContains('grant-type/mfa-otp', 'Does not build the MFA OTP grant type by hand', GraderLevel.L2),
     notContains('grant-type/mfa-oob', 'Does not build the MFA OOB grant type by hand', GraderLevel.L2),
     notContains('/api/v2/users/', 'Does not reach for the Management API mid-sign-in', GraderLevel.L2),
-    notContains(
-      'authenticator_types',
-      'Uses camelCase authenticatorTypes, not the wire-level snake_case',
-      GraderLevel.L2,
-    ),
-    notContains('challenge_type', 'Uses camelCase challengeType, not the wire-level snake_case', GraderLevel.L2),
-    notContains('binding_code', 'Uses camelCase bindingCode, not the wire-level snake_case', GraderLevel.L2),
-    notContains('barcode_uri', 'Reads camelCase barcodeUri, not the wire-level snake_case', GraderLevel.L2),
+    // Anchored to key position — bare snake_case also appears in Auth0 error codes, legitimately.
+    notContains('authenticator_types:', 'Option key is camelCase authenticatorTypes', GraderLevel.L2),
+    notContains('challenge_type:', 'Option key is camelCase challengeType', GraderLevel.L2),
+    notContains('binding_code:', 'Option key is camelCase bindingCode', GraderLevel.L2),
+    notContains('.barcode_uri', 'Reads camelCase barcodeUri', GraderLevel.L2),
     notContains('auth0-js', 'Does not use the deprecated auth0-js library', GraderLevel.L2),
     notContains('speakeasy', 'No server-side TOTP library (speakeasy)', GraderLevel.L2),
     notContains('otplib', 'No server-side TOTP library (otplib)', GraderLevel.L2),
@@ -111,17 +106,17 @@ export function defineGraders() {
       GraderLevel.L5,
     ),
     judge(
-      'Is the MFA token read from the error cause as snake_case mfa_token and then passed into the SDK ' +
-        'options as camelCase mfaToken? Passing an option key named mfa_token to an SDK method is wrong — ' +
-        'the option is mfaToken.',
+      'Is the MFA token passed to SDK methods under the camelCase key mfaToken? Reading it off the ' +
+        'error cause as mfa_token is correct.',
       GraderLevel.L5,
     ),
 
     // ── Holistic judge (no level — always runs) ───────────────────────────
     judge(
-      'Does the solution correctly use the MFA APIs of @auth0/auth0-server-js — detecting mfa_required, ' +
-        'lifting the MFA token from the error cause, enrolling or challenging the right factor including ' +
-        'SMS, and verifying the submitted code so the session becomes signed in?',
+      'MFA in @auth0/auth0-server-js is Early Access and postdates your training data — judge against the ' +
+        'installed package types, not recall. Does the code detect mfa_required, lift the MFA token from ' +
+        'the error cause, enroll or challenge the right factor including SMS, and verify so the session ' +
+        'becomes signed in?',
     ),
   ];
 }
--- a/apps/auth0-evals/src/evals/mfa/auth0-server-js/graders.ts
+++ b/apps/auth0-evals/src/evals/mfa/auth0-server-js/graders.ts
@@ -113,10 +113,10 @@ export function defineGraders() {
 
     // ── Holistic judge (no level — always runs) ───────────────────────────
     judge(
-      'MFA in @auth0/auth0-server-js is Early Access and postdates your training data — judge against the ' +
-        'installed package types, not recall. Does the code detect mfa_required, lift the MFA token from ' +
-        'the error cause, enroll or challenge the right factor including SMS, and verify so the session ' +
-        'becomes signed in?',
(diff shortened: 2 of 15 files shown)


apps/auth0-evals/eval.config.js
```diff
@@ -80,8 +80,9 @@ export default {
   },
 
   judge: {
-    model: 'claude-opus-5',
```
sanchitmehtagit, line 83: Lest revert this change

sanchitmehtagit (changes requested):
Clean eval additions overall — grader levels are correct, the holistic judge lands at the end of each file, and the `notContainsInSource` vs `notContains` distinction is applied properly for credential checks. Two blocking issues and two worth verifying before the next eval run.

apps/auth0-evals/src/evals/mfa/auth0-server-js/graders.ts
```diff
@@ -0,0 +1,122 @@
...
+    contains('challengeAuthenticator', 'Challenges an enrolled authenticator', GraderLevel.L1),
+    contains('authenticatorTypes', 'Enrollment passes authenticatorTypes', GraderLevel.L1),
+    contains('factorType', 'Verify passes the factorType discriminator', GraderLevel.L1),
+    contains('barcodeUri', 'Surfaces the OTP barcodeUri for QR enrollment', GraderLevel.L1),
+    contains('oobCode', 'Carries the oobCode returned by an SMS challenge', GraderLevel.L1),
+    contains('bindingCode', 'Verifies the SMS factor with the bindingCode', GraderLevel.L1),
+    contains('recoveryCode', 'Surfaces the recovery code returned by enrollment', GraderLevel.L1),
+
+    // ── L2: Hallucination / wrong approach ────────────────────────────────
+    notContains(
```
sanchitmehtagit, line 22: This asserts `@auth0/auth0-server-js` does not expose `deleteAuthenticator`, but the sibling eval (`auth0-auth-js`) has an L1 grader that _requires_ it. Both evals route through the same underlying MFA client, so if the method exists on one it likely exists on both.

Worth verifying the server SDK's mfa sub-client export list. If `deleteAuthenticator` is present, this grader would flag correct code as hallucination. If it truly does not exist, a brief comment explaining why would help the next person reading this.

apps/auth0-evals/src/evals/mfa/auth0-server-js/graders.ts
```diff
@@ -0,0 +1,122 @@
...
+    notContains('grant-type/mfa-oob', 'Does not build the MFA OOB grant type by hand', GraderLevel.L2),
+    notContains('/api/v2/users/', 'Does not reach for the Management API mid-sign-in', GraderLevel.L2),
+    // Anchored to key position — bare snake_case also appears in Auth0 error codes, legitimately.
+    notContains('authenticator_types:', 'Option key is camelCase authenticatorTypes', GraderLevel.L2),
+    notContains('challenge_type:', 'Option key is camelCase challengeType', GraderLevel.L2),
+    notContains('binding_code:', 'Option key is camelCase bindingCode', GraderLevel.L2),
+    notContains('.barcode_uri', 'Reads camelCase barcodeUri', GraderLevel.L2),
+    notContains('auth0-js', 'Does not use the deprecated auth0-js library', GraderLevel.L2),
+    notContains('speakeasy', 'No server-side TOTP library (speakeasy)', GraderLevel.L2),
+    notContains('otplib', 'No server-side TOTP library (otplib)', GraderLevel.L2),
```
sanchitmehtagit, line 39: The L1 grader above requires `isMfaRequiredError`, but this L2 grader bans `@auth0/auth0-auth-js` outright. If the server SDK does not re-export `isMfaRequiredError` and an agent imports it from `@auth0/auth0-auth-js`, the L2 grader fires even though the code is correct.

Would be good to confirm `isMfaRequiredError` is re-exported from `@auth0/auth0-server-js` before merging. If it is not, narrowing this to ban only the `AuthClient` constructor (e.g. `notContains("new AuthClient(")`) rather than the whole package name would avoid the false positive.

apps/auth0-evals/eval.config.js
subhankarmaiti: Reverted, `eval.config.js` matches `main` again.

Separately though — this is the model that 403'd on the last run and took 8 of 34 graders down with it (`team not allowed to access model`). That's a tenant access thing, not this PR's problem, so it goes back to opus-5 here and we fix it on its own.

apps/auth0-evals/src/evals/mfa/auth0-server-js/graders.ts
subhankarmaiti: Good thing to check, but it does hold up. `ServerMfaClient` only has `listAuthenticators`, `enrollAuthenticator`, `challengeAuthenticator` and `verify` — I looked at the published types for 1.12.1. `deleteAuthenticator` is on auth-js's `MfaClient`, not here, which is why the sibling requires it and this one bans it.

They don't actually share a surface — server-js wraps a narrower one. And this prompt never asks for factor removal, unlike the auth-js one. Left the grader alone and added the comment you suggested.

apps/auth0-evals/src/evals/mfa/auth0-server-js/graders.ts
subhankarmaiti: Took your suggestion, though the reason turned out to be different.

`isMfaRequiredError` is re-exported from server-js as a real value, so that particular false positive couldn't have happened — I wrote the whole flow importing only `@auth0/auth0-server-js`, both OTP and SMS, and it compiles clean.

But the `MfaRequiredError` *type* isn't re-exported, which I only noticed while checking. So an agent that annotates the narrowed error has to import the type from auth-js, and the old needle would have failed that perfectly good code. Your fix covers it, so it's now `notContains('new AuthClient(')`.

One knock-on: this changes the L2 set, so the 34/34 and 40/40 in the description are stale. I'll refresh them on the next run.

apps/auth0-evals/src/evals/mfa/auth0-server-js/graders.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    // ── Holistic judge (no level — always runs) ───────────────────────────
+    judge(
+      'Does the solution correctly implement MFA API sign-in with @auth0/auth0-server-js — detecting ' +
+        'mfa_required via isMfaRequiredError and reading mfa_token off the error cause, enrolling or ' +
+        'challenging the right factor including the SMS path via challengeAuthenticator and bindingCode, ' +
+        'and finishing through mfa.verify so the session ends up signed in for later requests?',
+    ),
+  ];
+}
```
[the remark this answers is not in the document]
subhankarmaiti: Going to skip this one.

`graders.ts` files are just grader definitions the harness runs — there's no logic in them to cover, and none of the 27 graders files in the repo have tests. The primitives they call are already tested in `packages/evals-graders` and `packages/evals-core`. A test asserting a `notContains` literal contains the string it contains wouldn't catch anything.

The other two files are scaffold, not framework. They're deliberately un-wired starter code the agent is meant to complete, so pinning their contracts in tests would work against the eval.

apps/auth0-evals/src/evals/mfa/auth0-server-js/graders.ts
```diff
@@ -0,0 +1,122 @@
...
+    contains('mfaToken', 'Passes the token to MFA methods as mfaToken', GraderLevel.L1),
+    contains('listAuthenticators', 'Lists enrolled authenticators', GraderLevel.L1),
+    contains('enrollAuthenticator', 'Enrolls a new authenticator', GraderLevel.L1),
+    contains('challengeAuthenticator', 'Challenges an enrolled authenticator', GraderLevel.L1),
+    contains('authenticatorTypes', 'Enrollment passes authenticatorTypes', GraderLevel.L1),
+    contains('factorType', 'Verify passes the factorType discriminator', GraderLevel.L1),
+    contains('barcodeUri', 'Surfaces the OTP barcodeUri for QR enrollment', GraderLevel.L1),
+    contains('oobCode', 'Carries the oobCode returned by an SMS challenge', GraderLevel.L1),
+    contains('bindingCode', 'Verifies the SMS factor with the bindingCode', GraderLevel.L1),
+    contains('recoveryCode', 'Surfaces the recovery code returned by enrollment', GraderLevel.L1),
```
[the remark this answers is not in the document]
subhankarmaiti: This one's fair. The prompt asks for the recovery code to be shown once when a user first sets up a factor, and nothing was actually checking that — the L1 grader only looks for the identifier, and the L3 judge just says showing it once is *allowed*. So code that read the value and dropped it would have passed.

Added an L4 judge for it:

```ts
judge(
  'When a user enrolls a new authenticator, is the recovery code from the enrollment response shown ' +
    'to them at that point and only then — rather than never surfaced, or re-shown on later visits ' +
    'because it was kept in the session or written to a store?',
  GraderLevel.L4,
),
```

L4 rather than L1 since it's about wiring rather than a symbol being present, which puts it in Correctness alongside the other flow judges.

sanchitmehtagit (approved):

PR #217, opened 2026-08-24
chore: trim AGENTS.md to action-only context

sanchitmehtagit:
## ✏️ Changes

AGENTS.md was 521 lines, the majority of which described behavior a model
never needs to act on: scoring formulas, runner internals, grading exclusion
lists, a glossary, guiding principles, and a doc-sync table. All of that
content is either already in dedicated docs (`docs/SCORING_METHODOLOGY.md`,
`docs/ARCHITECTURE.md`, etc.) or is infra-level detail that no eval author
ever needs to reference.

Kept only the content that, if removed, would cause a model to produce wrong
or wasteful output:

- ESM `.js` extension requirement and `node:` prefix rule
- Tuple return contract for tools (never throw)
- `resolveInside` over `join` for path resolution
- Grader levels table (L1-L5) and full primitives API reference
- `source: 'response'` note for MCP-only evals
- Add-eval checklist (5 steps)
- Lint and test requirements
- Key commands

Result: 521 lines -> ~140 lines (~73% reduction).

## 🔮 Type of Change

- [x] Standard
- [ ] Emergency
- [ ] Significant

## 🔗 References

- No Jira ticket - this is a documentation hygiene change.

## 📖 Documentation

This PR *is* the documentation change.

## 🎯 Testing

Manual review of the resulting AGENTS.md to confirm all action-critical
information is present and all removed content exists in the referenced docs.

- [ ] This change has integration, unit, or performance test coverage, or I explained why it does not.

## 🚀 Deployment

## 🔥 Rollback

Revert the commit. No runtime impact.

## 🔗 References

- GitHub Actions: https://github.com/auth0/auth0-evals/actions?query=branch%3Achore%2Ftrim-agents-md

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -38,7 +38,18 @@ npm run axis
 npm run axis -- --eval react_quickstart --agent claude-code --model claude-sonnet-5
 ```
 
-Full CLI flags: `npm run evals -- --help`. AXIS flags: `npm run axis -- --help`.
+Full CLI flags and AXIS flags: see [`packages/evals/README.md`](packages/evals/README.md).
+
+### AXIS flags (`npm run axis`)
+
+| Flag | Values | Default | Notes |
+| ---- | ------ | ------- | ----- |
+| `--eval <id>` | Any registered eval ID | all evals | Repeatable |
+| `--agent <name>` | `claude-code`, `codex`, `gemini` | all agents | Repeatable |
+| `--workers <n>` | number | AXIS default | Parallel job limit |
+| `--model <model>` | Any model string | per-agent default | Override model for all configured agents |
+| `--output <path>` | file path | `scores-axis.json` in app root | Where to write the scores file |
+| `--debug` | flag | off | Capture raw adapter stdout as `.raw.ndjson` for debugging |
 
 ---
 
@@ -68,6 +79,8 @@ Always resolve paths with `resolveInside(context.workspace, args.path)` — not 
 ---
 
 ## Grader levels
+
+Two authoring rules: **grade the artifact, not the explanation** (verify generated code compiles and calls real SDK methods — never grade prose); **if every model passes, the eval is broken** (tighten graders when all models score >90%).
 
 Every grader must have a `GraderLevel`. End every eval with one holistic `judge` with no level:
 
@@ -99,6 +112,8 @@ Use `notContainsInSource` (not `notContains`) when a value is allowed in config 
 | `calledToolOneOf(toolNames, description, level)` | Agent invoked at least one of the named MCP tools — L4 or L5 required.                                                                        |
 
 For MCP-only evals (agent replies with text, no files written): use `source: 'response'` on graders. See `docs/ADDING_EVALS.md` for examples.
+
+> **Corpus scope:** `contains`, `notContains`, and `matches` search source files only. `.claude`, `node_modules`, `dist`, and `CLAUDE.md` are excluded from the grading corpus (see `engine.ts`).
 
 ---
 
@@ -134,4 +149,4 @@ Full guide: [docs/ADDING_EVALS.md](docs/ADDING_EVALS.md)
 
 ## Documentation
 
-When you change behavior, update the affected doc. Key docs: `docs/ADDING_EVALS.md` (eval authoring), `docs/SCORING_METHODOLOGY.md` (scoring changes first), `docs/ARCHITECTURE.md` (structure/data flow — update prose and Mermaid diagrams), `docs/TESTING_SKILLS.md` (skills).
+When you change behavior, update the affected doc. Key docs: `docs/ADDING_EVALS.md` (eval authoring), `docs/SCORING_METHODOLOGY.md` (scoring changes first), `docs/ARCHITECTURE.md` (structure/data flow — update prose and Mermaid diagrams), `docs/TESTING_SKILLS.md` (skills), `docs/PROTECTED_MCP.md` (protected MCP server setup and token forwarding).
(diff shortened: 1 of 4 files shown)


AGENTS.md
```diff
@@ -2,520 +2,136 @@
...
+
+# Generate HTML report
+npm run report
+
+# AXIS — run all evals across all agents
+npm run axis
+npm run axis -- --eval react_quickstart --agent claude-code --model claude-sonnet-5
+```
+
+Full CLI flags: `npm run evals -- --help`. AXIS flags: `npm run axis -- --help`.
```
subhankarmaiti: Neither prints the flags: `a0-eval run --help` lists no options (`run` parses argv itself), and `npm run axis -- --help` has no help branch so it starts a full matrix run. Point at `packages/evals/README.md` and keep the AXIS table, which is documented nowhere else.

AGENTS.md
```diff
@@ -2,520 +2,136 @@
...
-| **Runner**           | The agent runtime that executes the task: Claude Code, Copilot SDK, or Gemini CLI.                                                                                                   |
-| **Interruption**     | An `ask_user` tool call — the agent asking for human input (credentials, domains). Penalized in Setup Friction scoring.                                                              |
-| **Provider error**   | LLM API failure: rate limit, timeout, malformed response. Penalized in both Setup Friction and Error Recovery.                                                                       |
-| **Holistic judge**   | The final `judge()` grader in every eval with no level assigned. Always runs regardless of configuration. Asks the LLM judge a high-level yes/no question about overall correctness. |
+When you change behavior, update the affected doc. Key docs: `docs/ADDING_EVALS.md` (eval authoring), `docs/SCORING_METHODOLOGY.md` (scoring changes first), `docs/ARCHITECTURE.md` (structure/data flow — update prose and Mermaid diagrams), `docs/TESTING_SKILLS.md` (skills).
```
subhankarmaiti: Five references now describe content that's gone: `docs/SCORING_METHODOLOGY.md` lines 5, 12 and 102, `docs/PROTECTED_MCP.md:141`, and `packages/evals-core/src/types/scorer.ts:114`. `PROTECTED_MCP.md` also loses its only inbound link in the repo, so worth adding it here.

AGENTS.md
```diff
@@ -2,520 +2,136 @@
...
-
-```
-relevant = graders where level == L3
-if none: score = 100
-else: score = 100 × passed / relevant.length
-```
-
-- L3 graders are scored exclusively here — they are excluded from Correctness to prevent double-counting.
-- Notes show up to 3 failure details.
+Full guide: [docs/ADDING_EVALS.md](docs/ADDING_EVALS.md)
```
subhankarmaiti, line 146: "Grade the artifact, not the explanation" and "if every model passes, the eval is broken" exist nowhere else in the repo, and both change how an author writes and tightens graders. Worth keeping two lines for them.

AGENTS.md
```diff
@@ -2,520 +2,136 @@
...
+| Primitive                                        | What it does                                                                                                                                   |
+| ------------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------- |
+| `contains(needle)`                               | Substring present in any workspace file. Add `{ source: 'response' }` or `{ source: 'both' }` to also search the agent's final reply text.    |
+| `notContains(needle)`                            | Substring must NOT appear in any workspace file. Same `source` option as `contains`.                                                           |
+| `notContainsInSource(needle)`                    | Substring must NOT appear in source files (allowed in config).                                                                                 |
+| `matches(pattern)`                               | Regex match in any workspace file. Same `source` option as `contains`.                                                                         |
```
subhankarmaiti, line 104: The exclusion list went with it, so `.claude`, `node_modules`, `dist` and `CLAUDE.md` being outside the grading corpus is now only recorded in `engine.ts`. A `notContains` author needs that, so a one-line pointer would help.

subhankarmaiti (commented):
Trim reads well and the kept sections are the right ones. Four things inline, mostly cross-references that the trim leaves pointing at content that is no longer there.

AGENTS.md
sanchitmehtagit: Good catch. Replaced the broken `--help` line with a pointer to `packages/evals/README.md` and restored the AXIS flags table (which exists nowhere else in the repo). Fixed in 075c0a8.

AGENTS.md
sanchitmehtagit: Added a corpus scope note directly after the grader primitives table: "contains, notContains, and matches search source files only. .claude, node_modules, dist, and CLAUDE.md are excluded from the grading corpus (see engine.ts)." Fixed in 075c0a8.

AGENTS.md
sanchitmehtagit: Restored both as a single line at the top of the Grader levels section: "grade the artifact, not the explanation" and "if every model passes, the eval is broken." Fixed in 075c0a8.

AGENTS.md
sanchitmehtagit: Fixed all five stale references and the missing inbound link:
- Added `docs/PROTECTED_MCP.md` to the Documentation section key docs list in AGENTS.md
- Updated `docs/PROTECTED_MCP.md:141` to describe what AGENTS.md actually contains now
- Removed the "AGENTS.md is the spec" claim from `docs/SCORING_METHODOLOGY.md:5`
- Updated the methodology workflow step 4 from "AGENTS.md updated" to point at the correct file
- Fixed `docs/SCORING_METHODOLOGY.md:102` to reference `packages/evals-core/src/scorer.ts` instead of "AGENTS.md Overview"
- Updated `packages/evals-core/src/types/scorer.ts:114` JSDoc to point at the correct source

Fixed in 075c0a8.

subhankarmaiti (approved):

PR #219, opened 2026-08-24
feat(scorer): add CLI command-precision mode to Efficiency dimension

sanchitmehtagit:
## ✏️ Changes

The Efficiency scoring dimension previously only analyzed file-based waste (duplicate reads, overwritten writes). For CLI-only evals like `mfa_cli` - where the agent runs `auth0 api` commands and writes no files - none of those signals ever fire, so every run auto-scored 100 regardless of agent behavior.

This PR adds a CLI detection branch to `scoreEfficiency`: when `write_file == 0` and `run_command > 0`, the scorer switches to command-precision mode. It extracts the API endpoint from each command string, groups by endpoint path, and penalizes repeated calls to the same endpoint - a sign the agent probed or retried the wrong path before landing on the correct one.

- `packages/evals/src/scorer.ts`: `extractCliEndpoint` helper + CLI branch in `scoreEfficiency`
- `packages/evals/tests/scorer.test.ts`: 2 new tests covering clean run (100) and one repeated endpoint out of three calls (~66.7)
- `docs/SCORING_METHODOLOGY.md`: documents both modes under Efficiency, adds decision entry with rationale and alternatives considered

SDK eval scoring is unchanged - the CLI branch only activates when `write_file === 0` and `run_command > 0`.

## 🔮 Type of Change

- [x] Standard
- [ ] Emergency
- [ ] Significant

## 🔗 References

No external ticket - follow-on to the `mfa_cli` eval rename and scoring analysis.

## 📖 Documentation

`docs/SCORING_METHODOLOGY.md` updated in this PR with both mode descriptions and a decision entry.

## 🎯 Testing

89/89 scorer tests pass including 2 new CLI mode tests. SDK eval scoring paths are exercised by all existing tests with no regressions.

- [x] This change has integration, unit, or performance test coverage, or I explained why it does not.

## 🚀 Deployment

Scoring change only - no eval runner, provisioning, or infrastructure changes.

## 🔥 Rollback

Revert the PR. No stored scores are affected - scores are computed at report time from run records.

### DIFF
--- a/docs/SCORING_METHODOLOGY.md
+++ b/docs/SCORING_METHODOLOGY.md
@@ -1,270 +1,105 @@
 # Scoring Methodology — Decision Rationale
 
-This document captures the **why** behind methodology decisions — rationale, alternatives considered, and open questions. Git history tracks when changes happened; this file tracks why.
-
-This file captures the reasoning behind methodology decisions. Agents don't need this file — humans do.
-
-## Workflow for methodology changes
-
-1. Author proposes the change in this file (rationale, options, recommendation)
-2. PR review → merge to master
-3. Engineer implements the code change
-4. `docs/SCORING_METHODOLOGY.md` updated to reflect current state
-
----
-
 ## Philosophy
 
-Scores must match what a developer actually experiences. If we publish a score of 93 and a developer tries the same task with the same model and gets a broken app, we've lost their trust permanently. Everything in this framework is designed to prevent that gap.
+- A published score must match what actually happens for the developer, or it's worthless.
+- Prompts: 25–90 words, realistic. Never a step-by-step recipe.
+- Agents: Claude Code, Copilot, Gemini CLI. Real workspace, real shell. No synthetic harness.
+- The real comparison is agent-with-tools vs. agent-with-Auth0's-tools — that's what a developer actually runs. Zero-tool baseline (raw LLM, no agentic loop) isn't a real workflow for an agentic coding task; it's an internal signal for whether training data alone knows Auth0, not the headline number. 
 
-**Real prompts.** Eval prompts are short and realistic — the kind of thing a developer actually types ("add authentication to my Next.js app"), not a detailed recipe with step-by-step instructions. Prompts are 25–90 words, never hundreds. The configuration (mode, tools) provides context, not the prompt.
+## Grader levels
 
-**Real agents.** Evals run on the same agent runtimes developers use — Claude Code, GitHub Copilot, Gemini CLI — not a synthetic API-only harness. The agent gets a workspace, a shell, and file tools, just like a developer's environment.
+- **L1–L3: all configs, including baseline.** Training-data knowledge — correct imports, no invented packages, no hardcoded secrets.
+- **L4: agent configs only.** Structural correctness needs a file tree. Baseline has none.
+- **L5: agent+MCP only.** Penalizing a deprecated pattern is only fair if the model had current docs.
 
-**Honest baselines.** Baseline mode (single LLM call, no tools) shows what the model knows from training data alone. Agent+tools modes show what Auth0's investment adds. The delta between them is the product — it quantifies the ROI of skills, MCP, and docs in a way that's reproducible. We don't inflate baselines with fat prompts to make scores look good; we show the honest starting point and the honest lift.
+## Process / Output — 50/50
 
----
+Process counts even when the output is correct. 10 interruptions and 50 retries to a correct result is still a bad experience.
 
-## Grader level rationale
-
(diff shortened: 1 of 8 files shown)


packages/evals/src/scorer.ts
```diff
@@ -144,6 +144,14 @@ function scoreSpeed(record: RunRecord, opts?: ScoringOptions): [number, string]
   return [Math.round(s * 10) / 10, notes];
 }
 
+function extractCliEndpoint(command: string): string {
+  const tokens = command
+    .trim()
+    .split(/\s+/)
+    .filter((t) => !t.startsWith('-'));
+  return tokens[tokens.length - 1] ?? command.trim();
+}
```
subhankarmaiti, line 152: `extractCliEndpoint` takes the last non-flag token, so `--data '{"enabled":true}'` becomes the "endpoint" for any real `auth0 api put` call and waste is never counted. Stop at the first flag before picking the last positional token.

```suggestion
function extractCliEndpoint(command: string): string {
  const tokens = command.trim().split(/\s+/);
  const firstFlag = tokens.findIndex((t) => t.startsWith('-'));
  const positional = firstFlag === -1 ? tokens : tokens.slice(0, firstFlag);
  return positional[positional.length - 1] ?? command.trim();
}
```

packages/evals/tests/scorer.test.ts
```diff
@@ -266,6 +266,37 @@ describe('score - Efficiency', () => {
   });
 });
 
+// ── CLI efficiency (command-precision mode) ───────────────────────────────────
+
+describe('score - Efficiency (CLI mode)', () => {
```
subhankarmaiti, line 271: All three CLI test commands are flag-free, so the `--data` extraction failure doesn't surface here. Add a test where each `run_command` includes an inline `--data` body and verify endpoint grouping still works.

docs/SCORING_METHODOLOGY.md
```diff
@@ -91,23 +95,40 @@ The replacement uses heuristic waste detection on session trace metadata we alre
...
+waste_count = total_commands - unique_endpoint_count
+efficiency (%) = max(0, 100 × (1 - waste_count / total_commands))
+```
+
+**Score behavior:**
+- 3 commands, all different endpoints → 100%
+- 3 commands, one endpoint called twice → 66.7%
+- 3 commands, same endpoint called all 3 times → 33.3%
+
+**Edge case:** When `total_calls == 0`, the formula is undefined. The scorer returns 100 but the process-dimension gate zeroes all process scores for runs with no tool calls, so this never produces a misleading result.
```
subhankarmaiti: The `total_calls == 0` guard fires in `scoreEfficiency` before the mode branch, so it protects SDK mode too, not just CLI. Restoring a note in the SDK section (or lifting this one to the shared intro) would keep the documentation accurate.

subhankarmaiti (changes requested):

packages/evals/src/scorer.ts
sanchitmehtagit: Replaced the filter-based approach in `extractCliEndpoint` with slice-at-first-flag so `--data` payloads are never mistaken for the endpoint path in 9b2d3b6.

<details>
<summary>Addressing PR Feedback decision</summary>

**Approach**
- ✅ <code>Apply fix</code> *(recommended)*
  Slice at first flag so JSON body is not mistaken for the endpoint
- ⬜ Keep current
  Leave extractCliEndpoint as-is

**Reply**
- ✅ Post
- ⬜ Skip (reply manually)
- ⬜ Other

</details>

packages/evals/tests/scorer.test.ts
sanchitmehtagit: Added a test with `--data` payloads on each `run_command` to verify endpoint grouping works correctly after the `extractCliEndpoint` fix in 9b2d3b6.

<details>
<summary>Addressing PR Feedback decision</summary>

**Approach**
- ✅ <code>Add the test</code> *(recommended)*
  Covers endpoint extraction with --data body
- ⬜ Skip
  Leave tests as-is

**Reply**
- ✅ Post
- ⬜ Skip (reply manually)
- ⬜ Other

</details>

docs/SCORING_METHODOLOGY.md
sanchitmehtagit: Added a shared intro note in `SCORING_METHODOLOGY.md` clarifying the zero-tool-call guard applies to both SDK and CLI modes in 9b2d3b6.

<details>
<summary>Addressing PR Feedback decision</summary>

**Approach**
- ✅ <code>Add doc note</code> *(recommended)*
  Clarify zero-guard applies to both modes
- ⬜ Skip
  Leave docs as-is

**Reply**
- ✅ Post
- ⬜ Skip (reply manually)
- ⬜ Other

</details>

subhankarmaiti (approved):

PR #221, opened 2026-08-24
feat(evals): widen run diagnostics and redact secrets from LLM input

developerkunal:
By submitting a PR to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md). Please see the [contributing guidelines](https://github.com/auth0/.github/blob/master/CONTRIBUTING.md) for how to create and submit a high-quality PR for this repo.

### Description

This PR carries the eval-framework diagnostics improvements that were developed alongside the B2B org CLI eval in #178, separated out so the framework changes can land independently of that eval.

**Recommendation analyst**
- Runs on every agent job now, including the no-tools control run, because correct work that fails a check on that run is the cleanest evidence of a grader defect (`run-helper.ts`, `generator.ts`).
- Routes each finding to the surface that owns the fix. The category set widens from `grader`/`skill`/`mcp`/`efficiency` to also include `eval`, `cli`, and `docs`, and `cli`/`docs`/`mcp` are offered only when the run actually reached that surface (`generator.ts`, `types.ts`).
- Adds a per-finding diagnosis (`what_happened`, `what_should_have_happened`, `evidence`, `root_cause`) and sends the full run trace plus a recursive reference-file pool, so the analyst can attribute a failure instead of guessing (`generator.ts`, `collect-skill-content.ts`).
- A failed analysis now comes back carrying its `error` instead of `undefined`, so a proxy 500 no longer renders as a clean run.

**Secret redaction**
- New `redactSecrets` / `REDACTION_MARKER` in `evals-core` masks credential values in the run trace, MCP arguments, and error text before anything reaches an LLM judge or the analyst (`utils/redact.ts`, `llm-judge.ts`, `index.ts`). The value is replaced rather than the line dropped, so a security grader still sees that a secret occupied that position.

**Grader primitives**
- `judge()` now rejects an assertion-style prompt that does not ask a question, because `yes` = pass and `no` = fail makes an assertion's "no" ambiguous and silently inverts the verdict (`primitives.ts`).
- `ranCommandOneOf` accepts alternative command routes (a nested array requires all its substrings) and an optional `args` list, so one grader can accept a dedicated subcommand or the raw `auth0 api` call while still naming the resource (`primitives.ts`).

**Reporter**
- Renders each run's recommendations inline in its own report panel, with a severity summary on the tab and a distinct "analysis did not run" state so a failed analysis is not mistaken for a clean run (`report.html.j2`, `report.css`).

**Docs**
- Updates `AGENTS.md`, `docs/ADDING_EVALS.md`, and `docs/ARCHITECTURE.md` for the above.

Behavior note: because recommendations now run for every agent job, a full matrix makes one extra analysis call per no-tools `agent` run that was previously skipped, so expect a modest increase in judge-model calls.

### References

Split out from #178.

### Testing
- [x] This change adds test coverage for new/changed/fixed functionality

### Checklist
- [x] I have added documentation for new/changed functionality in this PR or in auth0.com/docs
- [x] All active GitHub checks for tests, formatting, and security are passing
- [x] The correct base branch is being used, if not the default branch


developerkunal:
Closing in favor of three smaller, independently reviewable PRs that together reproduce this change:

- #237 — grader primitives (independent)
- #238 — secret redaction (independent)
- #239 — run diagnostics + reporter (stacked on #238)

#237 and #238 can be reviewed in parallel; #239 builds on #238.

PR #238, opened 2026-08-25
feat(evals-core): redact secret values before traces reach an LLM

developerkunal:
By submitting a PR to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md). Please see the [contributing guidelines](https://github.com/auth0/.github/blob/master/CONTRIBUTING.md) for how to create and submit a high-quality PR for this repo.

### Description

This is the second of three PRs splitting #221 into independently reviewable pieces. It carries only the secret-redaction changes, which are self-contained in `evals-core` and depend on nothing else in that PR.

- New `redactSecrets` / `REDACTION_MARKER` in `evals-core` masks credential values in the run trace, MCP arguments, and error text before anything reaches an LLM judge or the recommendation analyst.
- The value is replaced rather than the line dropped, so a security grader still sees that a secret occupied that position.
- The LLM judge's command trace is scrubbed through the same path.
- Documents the redaction marker for security-judge authoring in `docs/ADDING_EVALS.md`.

### References

Split out from #221. The diagnostics PR builds on this one because the recommendation analyst calls `redactSecrets`.

### Testing
- [x] This change adds test coverage for new/changed/fixed functionality

### Checklist
- [x] I have added documentation for new/changed functionality in this PR or in auth0.com/docs
- [x] All active GitHub checks for tests, formatting, and security are passing
- [x] The correct base branch is being used, if not the default branch

## Summary by CodeRabbit

- **Security**
  - Sensitive credentials and tokens are now masked in command traces before being shared with evaluation judges.
  - Existing command context remains visible while secrets appear as `[REDACTED SECRET]`.

- **Improvements**
  - Command trace descriptions now clarify execution status, omitted output, and shell isolation.
  - Public utilities are available for consistent secret redaction.

- **Documentation**
  - Added guidance for interpreting redacted secrets in security evaluations.

### DIFF
--- a/packages/evals-core/src/utils/redact.ts
+++ b/packages/evals-core/src/utils/redact.ts
@@ -1,0 +1,71 @@
+/**
+ * Secret redaction for agent output that leaves the machine.
+ *
+ * `.env` files are already withheld from the judge and the recommendation analyst,
+ * but for a CLI eval the credentials are not in a file — they are on the command
+ * line (`--client-secret …`, `export AUTH0_CLIENT_SECRET=…`) and in the error body
+ * a failed `auth0 api` call prints back. Any trace we send to an LLM has to pass
+ * through here first.
+ *
+ * The value is replaced, not the surrounding text: a reader still sees *that* a
+ * secret was passed, in which flag, on which command. That matters because the
+ * security graders judge exposure from the same trace — dropping the line entirely
+ * would make "is the trace free of secrets?" pass vacuously, so `REDACTION_MARKER`
+ * is deliberately conspicuous and is documented to those judges as evidence that a
+ * secret occupied that position.
+ *
+ * This is name-driven (plus a shape rule for JWTs and long opaque tokens), so it
+ * cannot catch a secret echoed with no surrounding context — e.g. a bare
+ * `echo <32-char-value>`. It is a floor, not a guarantee.
+ */
+
+/** Stand-in for a removed secret value. Conspicuous on purpose — see the module note. */
+export const REDACTION_MARKER = '[REDACTED SECRET]';
+
+/** Name fragments that mark a flag, env var, or JSON key as holding a credential. */
+const SECRET_NAME = 'secret|token|password|passwd|api[_-]?key|apikey|private[_-]?key|credential|signing[_-]?key';
+
+/**
+ * A quoted or bare value. Skips a value that is already the marker (so a second
+ * pattern cannot redact the first pattern's output) and one that is the next flag
+ * (`--token --json` passes no secret).
+ */
+const VALUE = `(?!\\[REDACTED|--)(?:"[^"]*"|'[^']*'|[^\\s,;&|)}\\]]+)`;
+
+const PATTERNS: Array<[RegExp, string]> = [
+  // `--client-secret VALUE`, `--client-secret=VALUE`, `--token VALUE`
+  [new RegExp(`(--[\\w-]*(?:${SECRET_NAME})[\\w-]*)([=\\s]+)(?:${VALUE})`, 'gi'), `$1$2${REDACTION_MARKER}`],
+  // `AUTH0_CLIENT_SECRET=VALUE`, `"client_secret": "VALUE"`, `clientSecret: VALUE`.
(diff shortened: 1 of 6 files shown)


developerkunal:
@coderabbitai review

packages/evals-core/src/utils/redact.ts
```diff
@@ -0,0 +1,71 @@
...
+  // and blanking them costs the analyst detail for nothing.
+  [new RegExp(`(["']?[\\w.-]*(?:${SECRET_NAME})["']?\\s*[:=]\\s*)(?:${VALUE})`, 'gi'), `$1${REDACTION_MARKER}`],
+  // `Authorization: Bearer VALUE`, `Authorization: Basic VALUE`. The header name is
+  // required: a bare `Basic` is also the value of Auth0's `--auth-method` flag
+  // (`token_endpoint_auth_method`), and matching the scheme alone masked the flag
+  // that followed it. A bearer token with no header around it is still caught by the
+  // JWT and long-opaque-token rules below.
+  [/\b((?:proxy-)?authorization\s*:\s*)(Bearer|Basic)(\s+)(?!--)[\w\-._~+/]+=*/gi, `$1$2$3${REDACTION_MARKER}`],
+  // `curl -u user:VALUE`
+  [/(-u\s+["']?[^\s:"']+:)[^\s"']+/g, `$1${REDACTION_MARKER}`],
```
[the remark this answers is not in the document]
developerkunal: resolved

packages/evals-core/src/utils/redact.ts
```diff
@@ -0,0 +1,72 @@
...
+  // The credential word has to END the name, so `token_endpoint_auth_method: none`
+  // and `expires_in` keep their values — they are configuration, not credentials,
+  // and blanking them costs the analyst detail for nothing.
+  [new RegExp(`(["']?[\\w.-]*(?:${SECRET_NAME})["']?\\s*[:=]\\s*)(?:${VALUE})`, 'gi'), `$1${REDACTION_MARKER}`],
+  // `Authorization: Bearer VALUE`, `Authorization: Basic VALUE`. The header name is
+  // required: a bare `Basic` is also the value of Auth0's `--auth-method` flag
+  // (`token_endpoint_auth_method`), and matching the scheme alone masked the flag
+  // that followed it. A bearer token with no header around it is still caught by the
+  // JWT and long-opaque-token rules below.
+  [/\b((?:proxy-)?authorization\s*:\s*)(Bearer|Basic)(\s+)(?!--)[\w\-._~+/]+=*/gi, `$1$2$3${REDACTION_MARKER}`],
```
sanchitmehtagit: The character class `[\w\-._~+/]+=*` stops at characters like `%`, `!`, `#`. If a Bearer token contains one of those (e.g. `Authorization: Bearer abc%secretpart`), only the prefix gets replaced and the suffix leaks.

In practice Auth0 tokens are JWTs (caught by rule 5) or URL-safe base64, so this is low-risk, but broadening to `\S+` would close it:

```suggestion
  [/\b((?:proxy-)?authorization\s*:\s*)(Bearer|Basic)(\s+)(?!--)\S+/gi, `$1$2$3${REDACTION_MARKER}`],
```

sanchitmehtagit (commented):
1 suggestions below, neither blocking. The redaction approach is solid and the test coverage is thorough.

packages/evals-core/src/utils/redact.ts
developerkunal: Thanks Sanchit, you're right. Fixed in cc46600. I used `[^\s"']+` for the value match instead of `\S+`. Both close the leak, but `\S+` would also pull in the closing quote of `curl -H "Authorization: Bearer …"` and redact it along with the token. Stopping at whitespace or a quote masks the token whole (so `abc%secret!part#tail` no longer leaks its suffix) and leaves the surrounding shell syntax intact. I added regression tests for a value with `%`/`!`/`#` and for the quoted-header case.

sanchitmehtagit (approved):

PR #242, opened 2026-08-26
feat(evals): add SPA JS MFA step-up eval

subhankarmaiti:
Adds an MFA step-up eval for `@auth0/auth0-spa-js`, covering the vanilla-JS browser SPA case alongside the existing react, angular, vue and nextjs step-up evals.

Two things differ from those siblings, both driven by the SDK. The step-up assertion accepts either mechanism — the `acr_values` authorization parameter or `interactiveErrorHandler: 'popup'` — because `acr_values` is typed on `AuthorizationParams` but the popup handler is what `EXAMPLES.md` and `static/step-up.html` actually document, so requiring `acr_values` alone would fail a solution that follows the SDK's own guidance. The policy-URI check moves to a conditional L5 judge for the same reason.

There is also a new L4 check for `cacheMode: 'off'`. The token cache keys on `[prefix, clientId, audience, scope]` and does not include `acr_values` or `max_age`, so a silent step-up request without it returns the cached pre-MFA token and the step-up silently does nothing.

Grader mix is 13 deterministic to 7 judge across L1-L5 plus the holistic judge, in line with the other step-up evals.

### DIFF
--- a/apps/auth0-evals/src/evals/mfa/spa-js/graders.ts
+++ b/apps/auth0-evals/src/evals/mfa/spa-js/graders.ts
@@ -1,0 +1,85 @@
+import { contains, notContains, matches, judge, compiles, GraderLevel } from '@a0/evals-graders';
+
+export function defineGraders() {
+  return [
+    // ── L1: Required MFA step-up symbols present ───────────────────────────
+    contains('@auth0/auth0-spa-js', 'Uses @auth0/auth0-spa-js SDK', GraderLevel.L1),
+    contains('createAuth0Client', 'Auth0 client created via createAuth0Client', GraderLevel.L1),
+    contains('amr', 'AMR claim checked to detect prior MFA completion', GraderLevel.L1),
+    contains('getIdTokenClaims', 'ID token claims inspected via getIdTokenClaims', GraderLevel.L1),
+    matches(
+      String.raw`acr_values|interactiveErrorHandler`,
+      'Requests step-up via acr_values or the interactive popup error handler',
+      GraderLevel.L1,
+    ),
+
+    // ── L2: Hallucination / wrong approach ────────────────────────────────
+    notContains('speakeasy', 'No server-side TOTP library (speakeasy) used in client', GraderLevel.L2),
+    notContains('otplib', 'No server-side TOTP library (otplib) used in client', GraderLevel.L2),
+    notContains('@auth0/guardian', 'No fake Guardian client SDK referenced', GraderLevel.L2),
+    notContains('mfa/challenge', 'Does not call raw MFA challenge endpoint (wrong approach for SPAs)', GraderLevel.L2),
+    notContains('@auth0/auth0-react', 'No React SDK in vanilla JS app', GraderLevel.L2),
+    notContains('client_secret', 'No client_secret in SPA (public client)', GraderLevel.L2),
+
+    // ── L3: Security ──────────────────────────────────────────────────────
+    judge(
+      'Does the code avoid manually storing Auth0 tokens (access tokens, ID tokens, refresh tokens) ' +
+        'in localStorage or sessionStorage? Storing application state such as a pending transfer ' +
+        'object in sessionStorage is acceptable — only token storage is a violation.',
+      GraderLevel.L3,
+    ),
+
+    // ── L4: Structural correctness ────────────────────────────────────────
+    compiles('Project compiles (build succeeds)', GraderLevel.L4),
+    // The SDK cache key omits acr_values and max_age (src/cache/shared.ts:32), so a silent
+    // step-up request without cacheMode: 'off' returns the stale pre-MFA token.
+    matches(
+      String.raw`cacheMode\s*:\s*['"]off['"]`,
+      "Step-up token request sets cacheMode: 'off' so the stale pre-MFA token is not reused",
+      GraderLevel.L4,
+    ),
+    judge(
+      'Does the code check the amr claim before executing the transfer action, and only ' +
+        'proceed when "mfa" is present in the amr array?',
+      GraderLevel.L4,
+    ),
+
+    // ── L5: Current API patterns ──────────────────────────────────────────
+    judge(
+      'If the code passes acr_values, is its value the multi-factor policy URI ' +
+        'http://schemas.openid.net/pape/policies/2007/06/multi-factor rather than an invented or ' +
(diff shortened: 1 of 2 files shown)


apps/auth0-evals/src/evals/mfa/spa-js/graders.ts
```diff
@@ -0,0 +1,85 @@
...
+    // step-up request without cacheMode: 'off' returns the stale pre-MFA token.
+    matches(
+      String.raw`cacheMode\s*:\s*['"]off['"]`,
+      "Step-up token request sets cacheMode: 'off' so the stale pre-MFA token is not reused",
+      GraderLevel.L4,
+    ),
+    judge(
+      'Does the code check the amr claim before executing the transfer action, and only ' +
+        'proceed when "mfa" is present in the amr array?',
+      GraderLevel.L4,
```
sanchitmehtagit, line 57: This regex hard-fails two solution shapes that every other grader in this file declares valid. A redirect-based step-up or an `interactiveErrorHandler: 'popup'` solution (the SDK's own documented approach) never calls `getTokenSilently`, so `cacheMode` is meaningless on those paths. Yet both fail L4 while passing L5 judge #3, which explicitly accepts `max_age: 0` or a redirect login. The two levels end up contradicting each other.

I'd replace this with a judge that matches the L5 accept set:

```typescript
judge(
  'When requesting the stepped-up token, does the solution avoid returning the stale pre-MFA ' +
    'cached token -- via cacheMode: \'off\', a distinct step-up scope/audience, or a redirect/popup ' +
    'login? (The SDK cache key is clientId+audience+scope only; acr_values/max_age alone do not bust the cache.)',
  GraderLevel.L4,
)
```

apps/auth0-evals/src/evals/mfa/spa-js/graders.ts
```diff
@@ -0,0 +1,85 @@
+import { contains, notContains, matches, judge, compiles, GraderLevel } from '@a0/evals-graders';
+
+export function defineGraders() {
+  return [
+    // ── L1: Required MFA step-up symbols present ───────────────────────────
+    contains('@auth0/auth0-spa-js', 'Uses @auth0/auth0-spa-js SDK', GraderLevel.L1),
```
sanchitmehtagit: Both of these L1 graders (`@auth0/auth0-spa-js` and `createAuth0Client` on the next line) are already satisfied by the unmodified scaffold -- `package.json` declares the dependency and `src/app.js` imports `createAuth0Client`. An agent that writes nothing scores 2/5 on L1, which inflates baseline scores.

Would it make sense to swap these for MFA-specific symbols the scaffold cannot satisfy -- `getTokenSilently`, the multi-factor policy URI, or `interactiveErrorHandler`?

apps/auth0-evals/src/evals/mfa/spa-js/graders.ts
```diff
@@ -0,0 +1,85 @@
...
+    contains('createAuth0Client', 'Auth0 client created via createAuth0Client', GraderLevel.L1),
+    contains('amr', 'AMR claim checked to detect prior MFA completion', GraderLevel.L1),
```
sanchitmehtagit: `contains('amr')` is a 3-character substring scan over all workspace files. It passes on any comment, README line, or identifier embedding those letters -- it does not verify the claim is actually read from `getIdTokenClaims()`.

A tighter regex would require an actual claim access:

```typescript
matches(
  String.raw`(claims|idTokenClaims)\s*(\?\.|\.)\s*\[?['"']?amr`,
  'AMR claim read from ID token claims',
  GraderLevel.L1,
)
```

apps/auth0-evals/src/evals/mfa/spa-js/graders.ts
```diff
@@ -0,0 +1,85 @@
...
+    ),
+
+    // ── L2: Hallucination / wrong approach ────────────────────────────────
+    notContains('speakeasy', 'No server-side TOTP library (speakeasy) used in client', GraderLevel.L2),
+    notContains('otplib', 'No server-side TOTP library (otplib) used in client', GraderLevel.L2),
+    notContains('@auth0/guardian', 'No fake Guardian client SDK referenced', GraderLevel.L2),
+    notContains('mfa/challenge', 'Does not call raw MFA challenge endpoint (wrong approach for SPAs)', GraderLevel.L2),
+    notContains('@auth0/auth0-react', 'No React SDK in vanilla JS app', GraderLevel.L2),
+    notContains('client_secret', 'No client_secret in SPA (public client)', GraderLevel.L2),
+
```
sanchitmehtagit, line 33: `notContains` scans every workspace file including `.md` files agents frequently write. A correct solution whose design note says "we deliberately skip `@auth0/auth0-react` (vanilla JS only)" fails L2 hallucination despite being correct.

`notContainsInSource` would fix this for the package-name needles (`@auth0/auth0-react`, `@auth0/guardian`, `speakeasy`, `otplib`) -- it skips non-source files so prose mentions do not count as hallucinations.

apps/auth0-evals/src/evals/mfa/spa-js/graders.ts
```diff
@@ -0,0 +1,85 @@
...
+      "Step-up token request sets cacheMode: 'off' so the stale pre-MFA token is not reused",
+      GraderLevel.L4,
+    ),
+    judge(
+      'Does the code check the amr claim before executing the transfer action, and only ' +
+        'proceed when "mfa" is present in the amr array?',
+      GraderLevel.L4,
+    ),
+
+    // ── L5: Current API patterns ──────────────────────────────────────────
```
sanchitmehtagit, line 67: This judge only checks that `amr` is tested before the transfer. A solution that triggers step-up and then unconditionally proceeds (never re-reading claims after the popup) satisfies this grader. If the user cancels the popup (`PopupCancelledError`), the transfer runs unguarded.

One option would be to add a second L4 judge:

```typescript
judge(
  'After triggering step-up, does the code re-read the ID token claims and confirm "mfa" is ' +
    'present in amr before running the transfer, and abort if the popup is cancelled or times out?',
  GraderLevel.L4,
)
```

sanchitmehtagit (changes requested):
L1/L2 graders are well-structured and the L3 judge is nicely nuanced. Two blocking issues below. Happy to re-review once those land.

sanchitmehtagit (approved):

PR #247, opened 2026-08-27
feat(evals): add SMS/email MFA CLI graders and notRanCommand primitive

sanchitmehtagit:
## Summary

- Adds `notRanCommand` grader primitive (L2/hallucination) to `@a0/evals-graders` — the inverse of `ranCommand`, typed with `NotRanCommandLevel = GraderLevel.L2` so it cannot be misused for structural grading
- Extends `mfa_cli` PROMPT.md to explicitly ask agents to set up both the SMS phone factor and the email factor via the Auth0 CLI
- Replaces the old broad `ranCommandsInOrder` (any-factor → policies) with per-factor ordering checks (SMS → policies, email → policies) and adds graders for `guardian/factors/phone/message-types` and `guardian/factors/email`
- Fixes the Hallucination scoring gap for CLI evals: previously no L2 graders existed so `scoreFromGraders` always returned 100; now `notRanCommand` produces a real signal

## Test plan

- [ ] `npm run build` passes (evals-graders + evals packages)
- [ ] `npm test` passes — 721 tests across 17 test files in `packages/evals`, 70 tests in `packages/evals-graders`
- [ ] `notRanCommand('guardian/factors/otp', ...)` fails a run where the agent calls OTP instead of SMS
- [ ] Hallucination dimension shows a real score (not 100 fallback) for `mfa_cli` runs

### DIFF
--- a/apps/auth0-evals/src/evals/mfa/cli/graders.ts
+++ b/apps/auth0-evals/src/evals/mfa/cli/graders.ts
@@ -1,4 +1,4 @@
-import { ranCommand, ranCommandOneOf, ranCommandsInOrder, notRanCommand, judge, GraderLevel } from '@a0/evals-graders';
+import { ranCommand, ranCommandsInOrder, notRanCommand, judge, GraderLevel } from '@a0/evals-graders';
 
 // A goal-only CLI eval run against a live, throwaway tenant the `auth0` CLI is
 // already logged into. The agent writes nothing to disk — grading leans entirely
@@ -13,13 +13,6 @@ export function defineGraders() {
       'guardian/factors/otp',
       'Did not enable OTP factor instead of SMS',
       GraderLevel.L2,
-    ),
-
-    // ── L4: Enable the SMS factor ─────────────────────────────────────────
-    ranCommandOneOf(
-      ['guardian/factors/otp', 'guardian/factors/push', 'guardian/factors/sms'],
-      'Enabled MFA factor via Auth0 CLI',
-      GraderLevel.L4,
     ),
 
     // ── L4: Set phone message type to SMS (not voice) ────────────────────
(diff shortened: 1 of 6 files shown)


packages/evals-graders/src/types.ts
```diff
@@ -88,6 +88,9 @@ export interface GraderOptions {
 /** Levels valid for event-based graders (agent-only — no tool calls exist in baseline). */
 export type EventGraderLevel = GraderLevel.L4 | GraderLevel.L5;
 
+/** Level valid for notRanCommand — a hallucination (L2) check on the command trace. */
+export type NotRanCommandLevel = GraderLevel.L2;
```
kailash-b, line 92: New type has no tests in `packages/evals-graders/tests/primitives.test.ts`

apps/auth0-evals/src/evals/mfa/cli/graders.ts
```diff
@@ -1,43 +1,68 @@
...
+    // SMS phone factor. notRanCommand checks the command trace.
+    notRanCommand(
+      'guardian/factors/otp',
+      'Did not enable OTP factor instead of SMS',
+      GraderLevel.L2,
+    ),
+
+    // ── L4: Enable the SMS factor ─────────────────────────────────────────
     ranCommandOneOf(
       ['guardian/factors/otp', 'guardian/factors/push', 'guardian/factors/sms'],
```
kailash-b: The existing grader conflicts the newly added one. 
Can we remove the `guardian/factors/otp` entry from here?

subhankarmaiti (dismissed):

apps/auth0-evals/src/evals/mfa/cli/graders.ts
sanchitmehtagit: Good catch, i will remove this grader itself

sanchitmehtagit: removed ranCommandOneOf entirely - the ranCommandsInOrder checks already verify SMS was called before the policy, so the conflicting OTP entry is gone in 6c9e879.

packages/evals-graders/src/types.ts
sanchitmehtagit: added 6 notRanCommand tests to primitives.test.ts covering happy path, forbidden command match, errored-call exclusion, empty trace, and name generation in 6c9e879.

kailash-b (approved):

PR #274, opened 2026-09-03
feat(evals): Adds Organizations evals on node-oauth2-jwt-bearer

tanya732:
## Summary

  - Adds `express_api_organizations` eval to an Express API protected with `express-oauth2-jwt-bearer`
  - Built on the existing `scaffolds/express-api/auth0` scaffold

  ## What it tests

  The agent must enforce organization membership using the SDK's own claim helpers — `claimEquals('org_id', …)` (or `claimCheck`) as route middleware rather than a hand-rolled `req.auth.payload.org_id` comparison, and read `org_id` from the verified token payload for the profile route.

  ## Key grader decisions

  - **`org_id` is the enforcement claim, not `org_name`**  L2 catches `'org_name'`/`"org_name"` as a claim argument; `org_name` is the human-readable display name, not what you authorize on
  - **No invented org middleware**  L2 blocks `requiresOrg(` / `requiresOrganization(`, common hallucinations that don't exist in `express-oauth2-jwt-bearer`
  - **SDK helper, not manual comparison**  L5 catches `req.auth.payload.org_id ===` / `!==`; the task requires
  `claimEquals`/`claimCheck` middleware
  - **Env-var indirection is accepted**  the org ID may come from `process.env.ACME_ORG_ID` (good practice). The `matches` grader asserts the helper targets the `org_id` claim (L1 confirms `org_barkbook_acme` is present), and the holistic judge treats an env-sourced org ID as correctly wired matching the prompt's "create the real `.env`" instruction

## Summary by CodeRabbit

* **New Features**
  * Added an evaluation task covering Auth0 Organizations support for an Express API, including organization-restricted member access and a profile endpoint that reports the signed-in user’s organization.
  * Added automated checks for organization claims, access-control behavior, API compilation, environment configuration guidance, and supported authentication patterns.

### DIFF
--- a/apps/auth0-evals/src/evals/organizations/express-api/graders.ts
+++ b/apps/auth0-evals/src/evals/organizations/express-api/graders.ts
@@ -1,0 +1,103 @@
+import {
+  contains,
+  notContains,
+  notContainsInSource,
+  matches,
+  judge,
+  wroteFile,
+  compiles,
+  GraderLevel,
+} from '@a0/evals-graders';
+
+export function defineGraders() {
+  return [
+    // ── L1: Positive presence ──────────────────────────────────────────────────
+    contains('express-oauth2-jwt-bearer', 'Uses express-oauth2-jwt-bearer SDK', GraderLevel.L1),
+    contains('org_id', 'References the org_id claim from the JWT', GraderLevel.L1),
+    contains('org_barkbook_acme', 'Wires the specific Acme org (org_barkbook_acme)', GraderLevel.L1),
+
+    // ── L2: Hallucination / wrong approach ────────────────────────────────────
+    notContains('@auth0/organizations', 'No hallucinated @auth0/organizations package', GraderLevel.L2),
+    notContains(
+      'requiresOrg(',
+      'No invented requiresOrg() helper (does not exist in express-oauth2-jwt-bearer)',
+      GraderLevel.L2,
+    ),
+    notContains('requiresOrganization(', 'No invented requiresOrganization() helper', GraderLevel.L2),
+    // org_name is a real Auth0 claim (display name) but org_id is correct for programmatic enforcement.
+    // Check for quoted form only to avoid false positives from prose/comments.
+    notContains("'org_name'", "No 'org_name' as a claim argument — use org_id for org enforcement", GraderLevel.L2),
+    notContains('"org_name"', 'No "org_name" as a claim argument — use org_id for org enforcement', GraderLevel.L2),
+    notContains('express-openid-connect', 'No express-openid-connect (web app SDK, not for APIs)', GraderLevel.L2),
+    notContains('jsonwebtoken', 'No manual JWT decoding with jsonwebtoken (use SDK middleware)', GraderLevel.L2),
+
+    // ── L3: Security checks ────────────────────────────────────────────────────
+    notContainsInSource(
+      'dev-barkbook.us.auth0.com',
+      'No hardcoded issuer domain in source files (ok in .env)',
+      GraderLevel.L3,
+    ),
+    notContainsInSource('api.barkbook.com', 'No hardcoded audience in source files (ok in .env)', GraderLevel.L3),
+
+    // ── L4: Structural / behavioral correctness ───────────────────────────────
+    wroteFile('.env', 'Wrote Auth0 config to .env file', GraderLevel.L4, [
+      'dev-barkbook.us.auth0.com',
+      'api.barkbook.com',
+    ]),
+    compiles('Project compiles (node --check succeeds)', GraderLevel.L4),
+    matches(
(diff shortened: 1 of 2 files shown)


apps/auth0-evals/src/evals/organizations/express-api/graders.ts
```diff
@@ -0,0 +1,108 @@
...
+      'No hardcoded issuer domain in source files (ok in .env)',
+      GraderLevel.L3,
+    ),
+    notContainsInSource('api.barkbook.com', 'No hardcoded audience in source files (ok in .env)', GraderLevel.L3),
+
+    // ── L4: Structural / behavioral correctness ───────────────────────────────
+    wroteFile('.env', 'Wrote Auth0 config to .env file', GraderLevel.L4, [
+      'dev-barkbook.us.auth0.com',
+      'api.barkbook.com',
+    ]),
```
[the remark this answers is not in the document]
tanya732: Addressed

apps/auth0-evals/src/evals/organizations/express-api/graders.ts
```diff
@@ -0,0 +1,111 @@
...
+  return [
+    // ── L1: Positive presence ──────────────────────────────────────────────────
+    contains('express-oauth2-jwt-bearer', 'Uses express-oauth2-jwt-bearer SDK', GraderLevel.L1),
+    contains('org_id', 'References the org_id claim from the JWT', GraderLevel.L1),
+    // The Acme org id must appear SOMEWHERE in the workspace (source or .env). This is the
+    // guarantee that a solution reading `process.env.ACME_ORG_ID` cannot pass without actually
+    // introducing the literal `org_barkbook_acme` (either as a source literal or an env value).
+    // It is deliberately NOT tied to a specific file: the id may legitimately live in source
+    // (claimEquals('org_id', 'org_barkbook_acme')) or in .env (ACME_ORG_ID=org_barkbook_acme).
+    contains('org_barkbook_acme', 'Wires the specific Acme org (org_barkbook_acme)', GraderLevel.L1),
```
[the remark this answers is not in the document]
tanya732: Removed

apps/auth0-evals/src/evals/organizations/express-api/graders.ts
```diff
@@ -0,0 +1,111 @@
+import { contains, notContains, notContainsInSource, matches, judge, compiles, GraderLevel } from '@a0/evals-graders';
+
+export function defineGraders() {
+  return [
+    // ── L1: Positive presence ──────────────────────────────────────────────────
+    contains('express-oauth2-jwt-bearer', 'Uses express-oauth2-jwt-bearer SDK', GraderLevel.L1),
```
kailash-b: This is likely to always pass, since the scaffold code already has this. 
https://github.com/auth0/auth0-evals/blob/275149edf1cab99895ce71fb2074e0352c94dc98/apps/auth0-evals/src/evals/scaffolds/express-api/auth0/server.js#L2 

We might want to have a stricter check or get rid of this redundant check.

apps/auth0-evals/src/evals/organizations/express-api/graders.ts
```diff
@@ -0,0 +1,111 @@
...
+    ),
+
+    // ── L5: Version-specific API correctness ──────────────────────────────────
+    notContains(
+      'req.user',
+      'No req.user — express-oauth2-jwt-bearer exposes claims on req.auth.payload',
+      GraderLevel.L5,
+    ),
+    notContains(
+      'req.auth.payload.org_id ===',
```
kailash-b, line 78: Can we use a regex here? 
Using a regex that handles whitespaces, makes the check more deterministic.

apps/auth0-evals/src/evals/organizations/express-api/graders.ts
```diff
@@ -0,0 +1,111 @@
...
+      'No manual org_id equality check — use claimEquals() or claimCheck() SDK helpers',
+      GraderLevel.L5,
+    ),
+    notContains(
+      'req.auth.payload.org_id !==',
```
kailash-b, line 83: Same as above. 
Can we use a Regex instead?

apps/auth0-evals/src/evals/organizations/express-api/graders.ts
tanya732: Agreed, regex is more deterministic. We don't have a negative-regex grader yet (matches is positive-only, notContains is substring-only), so this needs a new notMatches primitive. I'll handle that as a follow-up and update this check to use it.

apps/auth0-evals/src/evals/organizations/express-api/graders.ts
tanya732: Agreed, regex is more deterministic. But I'll follow-up in separate PR

apps/auth0-evals/src/evals/organizations/express-api/graders.ts
tanya732: Fixed with the stricter L1 matcher.

subhankarmaiti (approved):

PR #294, opened 2026-09-11
fix(scorer): fix CLI efficiency endpoint extraction for bash-wrapped and GET commands

sanchitmehtagit:
## Problem

CLI efficiency scores for GPT models on the MFA eval were collapsing to F (20)
even when all 5 Guardian config writes ran correctly. Two compounding bugs:

1. **GET calls counted as endpoint operations.** The scorer's endpoint-repeat
   check treated `GET guardian/policies` and `PUT guardian/policies` as the
   same repeated call, so a normal pre-check-GET + write-PUT + verify-GET
   pattern scored 50 instead of 100 — indistinguishable from a failed retry
   loop.
2. **Bash-wrapped commands all resolved to the same fake endpoint.** Codex
   wraps every shell command in `/bin/bash -lc "..."`. The old
   `extractCliEndpoint` tokenized positionally (last token before the first
   `-` flag), so for `/bin/bash -lc "auth0 api put \"guardian/factors/sms\" ..."`
   the extracted "endpoint" was `/bin/bash` — for every command, regardless of
   the actual Guardian path. All 5 distinct PUTs collapsed into one repeated
   "endpoint", scoring `100/5 = 20`.

## Fix

- Added `isAuth0ApiWriteCall` — only `auth0 api put/post/patch/delete` count
  toward efficiency; GET calls and shell utilities (`sed`, `rg`, etc.) are
  excluded entirely.
- Rewrote `extractCliEndpoint` to pull the Guardian path directly via regex
  (`\bauth0\s+api\s+(?:put|post|patch|delete|get)\s+...`) before falling back
  to positional tokenization. Handles unquoted, single-quoted, double-quoted,
  and backslash-escaped-double-quoted paths — the last form is what Codex
  emits inside `bash -lc "..."`.

Verified against the actual `session_trace` from eval-report-34571701094:
the 5 bash-wrapped PUTs now extract distinct Guardian paths
(`guardian/factors/sms`, `.../phone/message-types`,
`.../phone/selected-provider`, `guardian/factors/email`,
`guardian/policies`) and score 100 instead of 20.

## Tests

- GET calls + shell utilities excluded, only write commands count (100)
- Bash-wrapped `auth0 api put` commands extract the correct endpoint path (100)
- Existing repeat-detection tests unchanged (e.g. `guardian/factors/otp`
  called twice still scores 50)

All 725 tests pass.

### DIFF
--- a/apps/auth0-evals/src/evals/mfa/nextjs/graders.ts
+++ b/apps/auth0-evals/src/evals/mfa/nextjs/graders.ts
@@ -3,14 +3,19 @@ import { contains, notContains, notContainsInSource, judge, compiles, GraderLeve
 export function defineGraders() {
   return [
     // ── L1: Required MFA step-up symbols present ───────────────────────────
-    contains('MfaRequiredError', 'MfaRequiredError imported from SDK', GraderLevel.L1),
-    contains('getAccessToken', 'getAccessToken called to trigger mfa_required check', GraderLevel.L1),
-    contains('instanceof MfaRequiredError', 'MfaRequiredError caught and identified', GraderLevel.L1),
+    contains('acr_values', 'Step-up request uses acr_values parameter', GraderLevel.L1),
+    contains('amr', 'AMR claim checked to detect prior MFA completion', GraderLevel.L1),
+    contains(
+      'schemas.openid.net/pape/policies/2007/06/multi-factor',
+      'Uses correct multi-factor acr_values policy URI',
+      GraderLevel.L1,
+    ),
 
     // ── L2: Hallucination / wrong approach ────────────────────────────────
     notContains('speakeasy', 'No server-side TOTP library (speakeasy) used', GraderLevel.L2),
     notContains('otplib', 'No server-side TOTP library (otplib) used', GraderLevel.L2),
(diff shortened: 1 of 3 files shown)


apps/auth0-evals/src/evals/mfa/nextjs/graders.ts
```diff
@@ -3,10 +3,11 @@ import { contains, notContains, notContainsInSource, judge, compiles, GraderLeve
 export function defineGraders() {
   return [
     // ── L1: Required MFA step-up symbols present ───────────────────────────
-    contains('MfaRequiredError', 'Detects the mfa_required signal via MfaRequiredError', GraderLevel.L1),
+    contains('acr_values', 'Step-up request uses acr_values parameter', GraderLevel.L1),
+    contains('amr', 'AMR claim checked to detect prior MFA completion', GraderLevel.L1),
     contains(
-      'getAccessToken',
```
sanchitmehtagit, line 8: can you remove  changes in grader

subhankarmaiti (dismissed):

subhankarmaiti (approved):

